PR #12, opened 2022-04-21
Clang-format with pre-commit

ZenoTan:
Add pre-commit config. It automatically format existing cpp code.
.clang-format file now uses a simple chromium style.


rusty1s (approved):
LGTM.

PR #13, opened 2022-04-21
Initialize cpp test with `gtest` and github action

ZenoTan:


pyg_lib/test/cpp/test_cuda_version.cpp
```diff
@@ -0,0 +1,8 @@
+#include <gtest/gtest.h>
+
+#include <pyg/library.h>
+
+TEST(CudaVersionTest, BasicAssertions) {
+  // CPU detect no cuda
+  EXPECT_EQ(pyg::cuda_version(), (int64_t)-1);
```
yaoyaowd, line 7: I assume this test will fail on a gpu machine?

yaoyaowd (commented):
I did a little search on how other open source project organize repos with c++ code. I think we should change the directory a little bit to
```
cmake
pyg_lib
  csrc
    // different packages.
    // for each package, we can put *.h, *.cc together.
    pyg
      library.h
      library.cc
test
```
I think it is similar to how pytorch organize their code. WDYT?

rusty1s:
@yaoyaowd I tried to follow https://github.com/pytorch/vision/tree/main/torchvision/csrc structure.

yaoyaowd:
Yes, I think what you originally planned make sense to me. This PR changed the organization a little bit. We can move cmake and test to top level directory. And we can put cc file and h file together in the csr directory or different packages in csr directory.
I see separate header in include directory in other projects but I think for this one we should follow the pytorch family style.

ZenoTan:
> Yes, I think what you originally planned make sense to me. This PR changed the organization a little bit. We can move cmake and test to top level directory. And we can put cc file and h file together in the csr directory or different packages in csr directory. I see separate header in include directory in other projects but I think for this one we should follow the pytorch family style.

I will try to follow the torch style

pyg_lib/test/cpp/test_cuda_version.cpp
ZenoTan: I will add another test flag to specify if cuda is enabled for the test

CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ set(PYG_VERSION 0.0.0)
 
 option(WITH_CUDA "Enable CUDA support" OFF)
 option(USE_PYTHON "Link to Python when building" ON)
+option(BUILD_TEST "Enable testing" ON)
+option(TEST_WITH_CUDA "CUDA testing" OFF)
```
rusty1s, line 9: Do we need this here? IMO, `WITH_CUDA` is already sufficient.

CMakeLists.txt
```diff
@@ -31,3 +35,11 @@ endif()
 set_target_properties(${PROJECT_NAME} PROPERTIES
   EXPORT_NAME PyG
   INSTALL_RPATH ${TORCH_INSTALL_PREFIX}/lib)
+
+IF(TEST_WITH_CUDA)
+  add_definitions(-DTEST_WITH_CUDA)
+ENDIF()
+
+IF(BUILD_TEST)
```
rusty1s, line 43:
```suggestion
if(BUILD_TEST)
```
Unify syntax

CMakeLists.txt
```diff
@@ -31,3 +35,11 @@ endif()
...
+  INCLUDE(cmake/test.cmake)
+ENDIF()
```
rusty1s, line 45:
```suggestion
endif()
```

CMakeLists.txt
```diff
@@ -31,3 +35,11 @@ endif()
 set_target_properties(${PROJECT_NAME} PROPERTIES
   EXPORT_NAME PyG
   INSTALL_RPATH ${TORCH_INSTALL_PREFIX}/lib)
+
+IF(TEST_WITH_CUDA)
```
rusty1s, line 39: I think this is safe to drop. Let me know if you disagree.

test/cpp/test_cuda_version.cpp
```diff
@@ -0,0 +1,11 @@
+#include <gtest/gtest.h>
+
+#include "../../pyg_lib/csrc/library.h"
+
+TEST(CudaVersionTest, BasicAssertions) {
+#ifdef TEST_WITH_CUDA
```
rusty1s, line 6:
```suggestion
#ifdef WITH_CUDA
```

test/cpp/test_cuda_version.cpp
```diff
@@ -0,0 +1,11 @@
...
+  EXPECT_NE(pyg::cuda_version(), (int64_t)-1);
```
rusty1s, line 7: Do we really need the `(int64_t)` cast here? Man, I hate C++.

rusty1s (approved):

rusty1s (commented):
Can we also add some quick documentation on how to build and test in `CONTRIBUTING.md`? Just a few code lines are sufficient.

cmake/test.cmake
```diff
@@ -0,0 +1,21 @@
...
+  googletest
+  URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
+)
+# For Windows: Prevent overriding the parent project's compiler/linker settings
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
+
+
+enable_testing()
+FILE(GLOB cpp_tests pyg_lib/test/cpp/test_*.cpp)
```
rusty1s, line 12:
```suggestion
FILE(GLOB tests test/csrc/**/test_*.cpp)
```
Let's move tests to the root dir.

rusty1s:
It looks like we need to move the test file now (no tests found reported :))

test/cpp/test_cuda_version.cpp
ZenoTan: Removed it and found it still worked :)

PR #14, opened 2022-04-22
Python `CMakeExtension`

rusty1s:
Allows `cmake` build via `pip install .`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 import os
 import os.path as osp
 import subprocess
-import sys
 
 import torch
 from setuptools import Extension, find_packages, setup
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -1,8 +1,59 @@
...
+
+        extdir = os.path.abspath(osp.dirname(self.get_ext_fullpath(ext.name)))
+
+        if self.debug is None:
+            self.debug = bool(int(os.environ.get('DEBUG', 0)))
+
+        if not osp.exists(self.build_temp):
+            os.makedirs(self.build_temp)
+
+        cmake_args = [
```
ZenoTan, line 38: Do we need to disable `BUILD_TEST` for just `pip install`

setup.py
```diff
@@ -1,8 +1,59 @@
...
+            '-DUSE_PYTHON=ON',
+            f'-DCMAKE_LIBRARY_OUTPUT_DIRECTORY={extdir}',
+            f'-DCMAKE_BUILD_TYPE={"DEBUG" if self.debug else "RELEASE"}',
+            f'-DCMAKE_PREFIX_PATH={torch.utils.cmake_prefix_path}',
+            f'-DWITH_CUDA={"ON" if torch.cuda.is_available() else "OFF"}',
```
ZenoTan, line 43:
```suggestion
            f'-DWITH_CUDA={"ON" if torch.cuda.is_available() else "OFF"}',
            '-DBUILD_TEST=OFF',
```

setup.py
```diff
@@ -36,5 +87,7 @@
         'dev': dev_requires,
     },
     packages=find_packages(),
+    ext_modules=[CMakeExtension('libpyg')],
```
ZenoTan, line 90: A little confused about naming it `libpyg` but in python having `torch.ops.pyg`, and folder name `pyg_lib`. Is there a way to make them look consistent?

setup.py
rusty1s: Yes, thank you :)

setup.py
rusty1s: I currently do not see an easy way to do this.
* `lib*.so` is the standard naming of libraries
* `pyg` is the namespace
* `pyg_lib` is the Python package name.

I added a TODO here which we can look into later. Hope that works for you :)

PR #25, opened 2022-04-26
`random_walk` CUDA implementation

rusty1s:


rusty1s:
First CUDA kernel implementation in `pyg-lib` :) Let me know what you think!

pyg_lib/csrc/sampler/cuda/random_walk_kernel.cu
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+at::Tensor random_walk_kernel(const at::Tensor& rowptr,
+                              const at::Tensor& col,
+                              const at::Tensor& seed,
+                              int64_t walk_length,
+                              double p,
+                              double q) {
+  TORCH_CHECK(rowptr.is_cuda(), "'rowptr' must be a CUDA tensor");
```
ZenoTan, line 49: I'm curious that: will device information be checked when torch dispatch CPU/GPU kernels?

rusty1s:
> IMO we should put cpp and cuda tests separately and test both. This also applies to test utils such as `graph.h`

Sounds good. I will do this in a follow-up PR.

PR #26, opened 2022-04-26
`randint` utils for samplers

ZenoTan:
Added `randint` to enable fast random. It can be integrated into random samplers.

### DIFF
--- a/pyg_lib/csrc/random/cpu/randint_engine.h
+++ b/pyg_lib/csrc/random/cpu/randint_engine.h
@@ -12,6 +12,13 @@ const int RAND_PREFETCH_THRESHOLD = 128;
 // Use torch::randint to generate 64-bit random numbers
 const int RAND_PREFETCH_BITS = 64;
 
+/**
+ * Amortized O(1) uniform random within a range.
+ *
+ * Reduced torch random overheads by prefetching large random bits in advance.
+ * Prefetched random bits are consumed on demand.
+ *
+ */
 class PrefetchedRandint {
  public:
   PrefetchedRandint()
@@ -67,6 +74,10 @@ class PrefetchedRandint {
   int bits_;
 };
 
+/**
+ * Randint functor for uniform integer distribution.
+ * Wrapped PrefetchedRandint as its efficient core implementation.
+ */
 template <typename T>
 class RandintEngine {
  public:
--- a/pyg_lib/csrc/random/cpu/randint_engine.h
+++ b/pyg_lib/csrc/random/cpu/randint_engine.h
@@ -14,9 +14,10 @@ const int RAND_PREFETCH_BITS = 64;
 
 class PrefetchedRandint {
  public:
-  PrefetchedRandint(int size = RAND_PREFETCH_THRESHOLD,
-                    int bits = RAND_PREFETCH_BITS) {
-    prefetch_();
+  PrefetchedRandint()
+      : PrefetchedRandint(RAND_PREFETCH_THRESHOLD, RAND_PREFETCH_BITS) {}
+  PrefetchedRandint(int size, int bits) : size_(size), bits_(bits) {
+    prefetch(size_);
   }
 
   template <typename T>
@@ -24,19 +25,19 @@ class PrefetchedRandint {
     unsigned needed = 64;
 
     // Mutiple levels of range to save prefetched bits
-    if (range <= (1 << 15)) {
+    if (range < (1 << 16)) {
       needed = 16;
-    } else if (range <= (1UL << 31)) {
+    } else if (range < (1UL << 32)) {
       needed = 32;
     }
 
     if (bits_ < needed) {
-      if (size_ > 0) {
+      if (size_ > 1) {
         size_--;
         bits_ = RAND_PREFETCH_BITS;
       } else {
         // Prefetch if no enough bits
-        prefetch_();
+        prefetch(prefetched_randint_.size(0));
       }
     }
 
@@ -44,7 +45,7 @@ class PrefetchedRandint {
     uint64_t* prefetch_ptr =
         reinterpret_cast<uint64_t*>(prefetched_randint_.data_ptr<int64_t>());
     uint64_t mask = (needed == 64) ? std::numeric_limits<uint64_t>::max()
-                                   : (1ULL << needed) - 1;
+                                   : ((1ULL << needed) - 1);
     uint64_t res = (prefetch_ptr[size_ - 1] & mask) % range;
     prefetch_ptr[size_ - 1] >>= needed;
     bits_ -= needed;
@@ -52,12 +53,12 @@ class PrefetchedRandint {
   }
 
  private:
-  void prefetch_() {
-    prefetched_randint_ = torch::randint(
-        std::numeric_limits<int64_t>::min(),
-        std::numeric_limits<int64_t>::max(), {RAND_PREFETCH_THRESHOLD},
-        torch::TensorOptions().dtype(torch::kInt64));
-    size_ = RAND_PREFETCH_THRESHOLD;
+  void prefetch(int size) {
+    prefetched_randint_ =
+        torch::randint(std::numeric_limits<int64_t>::min(),
+                       std::numeric_limits<int64_t>::max(), {size},
+                       torch::TensorOptions().dtype(torch::kInt64));
+    size_ = size;
     bits_ = RAND_PREFETCH_BITS;
   }
 
--- a/pyg_lib/csrc/random/cpu/randint_engine.h
+++ b/pyg_lib/csrc/random/cpu/randint_engine.h
@@ -1,0 +1,55 @@
+#include <torch/torch.h>
+
+#include <limits.h>
+
+namespace pyg {
+namespace random {
+const int RAND_PREFETCH_THRESHOLD = 128;
+const int RAND_PREFETCH_BITS = 64;
+template <typename T>
+class RandintEngine {
+ public:
+  RandintEngine() : size_(RAND_PREFETCH_THRESHOLD), bits_(64) {
+    prefetch_randint_ = torch::randint(
+        0L, std::numeric_limits<int64_t>::max(), {RAND_PREFETCH_THRESHOLD},
+        torch::TensorOptions().dtype(torch::kInt64));
+  }
+  template <unsigned B>
+  T rand(T range) {
+    T num;
+    if (bits_ < B) {
+      if (size_ > 0) {
+        size_--;
+        bits_ = 64;
+      } else {
+        prefetch_randint_ = torch::randint(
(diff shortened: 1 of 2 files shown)


pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,55 @@
+#include <torch/torch.h>
+
+#include <limits.h>
+
+namespace pyg {
+namespace random {
+const int RAND_PREFETCH_THRESHOLD = 128;
+const int RAND_PREFETCH_BITS = 64;
```
yaoyaowd, line 8: nit: empty line.
```suggestion
const int RAND_PREFETCH_BITS = 64;

template <typename T>
```

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,55 @@
...
+template <typename T>
+class RandintEngine {
+ public:
+  RandintEngine() : size_(RAND_PREFETCH_THRESHOLD), bits_(64) {
+    prefetch_randint_ = torch::randint(
+        0L, std::numeric_limits<int64_t>::max(), {RAND_PREFETCH_THRESHOLD},
+        torch::TensorOptions().dtype(torch::kInt64));
+  }
```
yaoyaowd, line 16: nit: empty line
```suggestion
  }
  
  template <unsigned B>
```

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,55 @@
+#include <torch/torch.h>
```
yaoyaowd, line 1: `#pragma once`

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,55 @@
...
+ public:
+  RandintEngine() : size_(RAND_PREFETCH_THRESHOLD), bits_(64) {
+    prefetch_randint_ = torch::randint(
+        0L, std::numeric_limits<int64_t>::max(), {RAND_PREFETCH_THRESHOLD},
+        torch::TensorOptions().dtype(torch::kInt64));
+  }
+  template <unsigned B>
+  T rand(T range) {
+    T num;
+    if (bits_ < B) {
```
yaoyaowd, line 20: seems it never happen.

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,55 @@
...
+        bits_ = 64;
+      } else {
+        prefetch_randint_ = torch::randint(
+            0L, std::numeric_limits<int64_t>::max(), {RAND_PREFETCH_THRESHOLD},
+            torch::TensorOptions().dtype(torch::kInt64));
+        size_ = RAND_PREFETCH_THRESHOLD;
+        bits_ = RAND_PREFETCH_BITS;
+      }
+    }
+    int64_t* prefetch_ptr = prefetch_randint_.data_ptr<int64_t>();
```
yaoyaowd, line 32: we can define this as a class variable and reset it whenever we reset `prefetch_randint_`.

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,55 @@
...
+    int64_t res = (prefetch_ptr[size_ - 1] % range) & ((1ULL << B) - 1);
```
yaoyaowd, line 33: why do we want `((1ULL << B) - 1);`? to make sure we return positive numbers? I think maybe we can make sure whatever in prefetch_ptr is unsigned.

yaoyaowd (commented):
Could you add some document for the class as well?
Also by eyeball the code, `bits_` seems always 64, so it is larger than B and never trigger the code in the if condition.

pyg_lib/csrc/random/cpu/randint_engine.h
ZenoTan: I've fixed it to the expected logic

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,88 @@
...
+    uint64_t mask = (needed == 64) ? std::numeric_limits<uint64_t>::max()
+                                   : (1ULL << needed) - 1;
+    uint64_t res = (prefetch_ptr[size_ - 1] & mask) % range;
+    prefetch_ptr[size_ - 1] >>= needed;
+    bits_ -= needed;
+    return (T)res;
+  }
+
+ private:
+  void prefetch_() {
```
yaoyaowd, line 55: For functions, we don't need suffix `_` for functions in a class. also
```
void prefetch(int size, int bits)
```

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,88 @@
...
+const int RAND_PREFETCH_THRESHOLD = 128;
+
+// Use torch::randint to generate 64-bit random numbers
+const int RAND_PREFETCH_BITS = 64;
+
+class PrefetchedRandint {
+ public:
+  PrefetchedRandint(int size = RAND_PREFETCH_THRESHOLD,
+                    int bits = RAND_PREFETCH_BITS) {
+    prefetch_();
```
yaoyaowd, line 19:
```suggestion
    prefetch_(size, bits);
```

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,88 @@
...
+namespace random {
+
+const int RAND_PREFETCH_THRESHOLD = 128;
+
+// Use torch::randint to generate 64-bit random numbers
+const int RAND_PREFETCH_BITS = 64;
+
+class PrefetchedRandint {
+ public:
+  PrefetchedRandint(int size = RAND_PREFETCH_THRESHOLD,
```
yaoyaowd, line 17: I think a preferred way of define constructors is
```
PrefetchedRandint() {
  this(RAND_PREFETCH_THRESHOLD, RAND_PREFETCH_BITS)
}

PrefetchRandint(int size, int bits) {
  prefetch(size, bits)
}
```

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,88 @@
...
+        prefetch_();
+      }
+    }
+
+    // Currently torch could only make 64-bit signed numbers
+    uint64_t* prefetch_ptr =
+        reinterpret_cast<uint64_t*>(prefetched_randint_.data_ptr<int64_t>());
+    uint64_t mask = (needed == 64) ? std::numeric_limits<uint64_t>::max()
+                                   : (1ULL << needed) - 1;
+    uint64_t res = (prefetch_ptr[size_ - 1] & mask) % range;
```
yaoyaowd, line 48: I feel if `needed=16`, and `bits_=64`, the next 4 calls of `rand(1<<15)` will return the same number because `size_` is the same, `mask` is the same, and `range` is the same.

ZenoTan: The prefetched element is shifted accordingly after that.

test/csrc/random/test_randint.cpp
```diff
@@ -0,0 +1,79 @@
+#include <gtest/gtest.h>
+
+#include <vector>
+
+#include "../../../pyg_lib/csrc/random/cpu/randint_engine.h"
+
```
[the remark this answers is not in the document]
ZenoTan: Added one. It sometimes failed for 10000 times so I just have 1000 times. You have made a good guess :)

pyg_lib/csrc/random/cpu/randint_engine.h
ZenoTan: Treat random numbers as just unsigned bits by pointer casting

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,100 @@
+#pragma once
+
+#include <torch/torch.h>
+
+#include <limits.h>
+
+namespace pyg {
+namespace random {
+
+const int RAND_PREFETCH_THRESHOLD = 128;
```
yaoyaowd, line 10: nit: I think this should be called RAND_PREFETCH_SIZE instead of THRESHOLD.

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * Reduced torch random overheads by prefetching large random bits in advance.
+ * Prefetched random bits are consumed on demand.
+ *
+ */
+class PrefetchedRandint {
+ public:
+  PrefetchedRandint()
+      : PrefetchedRandint(RAND_PREFETCH_THRESHOLD, RAND_PREFETCH_BITS) {}
+  PrefetchedRandint(int size, int bits) : size_(size), bits_(bits) {
```
yaoyaowd, line 26: I think we don't need `size_(size), bits_(bits)` since they were initialized in `prefetch` function.

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,100 @@
...
+    return (T)res;
+  }
+
+ private:
+  void prefetch(int size) {
+    prefetched_randint_ =
+        torch::randint(std::numeric_limits<int64_t>::min(),
+                       std::numeric_limits<int64_t>::max(), {size},
+                       torch::TensorOptions().dtype(torch::kInt64));
+    size_ = size;
```
yaoyaowd, line 68: thoughts, what if we do `size_=size - 1` and in the function above you can directly use `prefetch_ptr[size]` instead of `size-1` everywhere.

yaoyaowd (approved):
A few minor comments. The idea of using 64 bit for buffer and get 4 of 16 bit random number is brilliant. Is this benchmarked?

rusty1s (commented):
Overall, this looks good to me, and we can definitely merge this in. I feel that `PrefetchedRandint` needs more documentation and comments though.

Any reason why we go with this prefetching approach in the first place? It seems hard to extend if we think about weighted sampling in later stages. What speaks against using `CPUGeneratorImpl` directly from PyTorch?
```cpp
CPUGeneratorImpl* generator = getDefaultCPUGenerator();
generator->random()
```

pyg_lib/csrc/random/cpu/randint_engine.h
```diff
@@ -0,0 +1,118 @@
+#pragma once
+
+#include <torch/torch.h>
```
rusty1s, line 3:
```suggestion
#include <ATen/Aten.h>
```
and replace all `torch::` with `at::`. Just found out that this heavily improves compilation time.

PR #27, opened 2022-04-27
Update build instructions

rusty1s:
Added `ctest` and `ninja` information.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 ## Build & Test
 
-`pyg-lib` can be build as a standalone C++/CUDA library.
+`pyg-lib` can be built as a standalone C++/CUDA library.
 For this, we adopt [CMake](https://cmake.org/) to build and [GoogleTest](https://github.com/google/googletest) to test.
 
 First, install all requirements:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,8 +1,28 @@
...
+```
 
-You can also build `pyg_lib` as a Python library using `pip install .` with our C++ library as a `CMakeExtension`, which will be loaded (`libpyg.so`) when you `import pyg_lib`
+Then, build the library via:
+
+```
+mkdir build
+cd build
+export Torch_DIR=`python -c 'import torch;print(torch.utils.cmake_prefix_path)'`
+cmake .. -GNinja -DBUILD_TEST=ON -DWITH_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=75
```
ZenoTan, line 18: CUDA arch could be configured by user variable than a single number

ZenoTan (approved):

CONTRIBUTING.md
ZenoTan: Like in `torch_geometric` just have a `${CUDA_ARCH}` and let users set that, or rely on CMake to auto detect (which IMO is more complicated)

PR #30, opened 2022-04-27
explicitly specify where to import headers

yaoyaowd:
In most C++ code style, we are not recommend to use relative path in include path because it is fragile to refactor and hard to debug when `../../blabla/../../blabla` happens.


ZenoTan:
LGTM.
Wondering is there a reason for not using `"xx.h"` instead of `<xx.h>` even we are not doing relative import?

ZenoTan (approved):

yaoyaowd:
It is designed to distinguish current working directory vs system headers. By working directory, it is not the file's directory but `pyg_lib`.

rusty1s (approved):
Thanks!

PR #38, opened 2022-04-30
Weighted graph preprocessing to enable biased sampling

ZenoTan:


pyg_lib/csrc/random/cpu/biased_sampling.h
```diff
@@ -127,8 +127,166 @@ index_t biased_random_alias(const index_t* idx,
...
+ * each entry, we can either sample the entry index itself or the alias index.
+ *
+ * The output bias is grouped by the neighbors of each node. For
+ * each group of neighbors, the number (p) means the probability of sampling the
+ * index of this alias table entry instead of its alias index (1 - p).
+ *
+ * Example:
+ *
+ * Neighbors of a node has the following bias: {0.5, 3, 0.5}
+ * The output bias of this group will be: {0.5, 1.0, 0.5}
```
yaoyaowd, line 212: I think here should be something like {0.375, 1, 0.375}?

pyg_lib/csrc/random/cpu/biased_sampling.h
```diff
@@ -127,8 +127,166 @@ index_t biased_random_alias(const index_t* idx,
...
+template <typename scalar_t>
+void biased_to_alias_helper(int64_t* rowptr_data,
+                            int64_t rowptr_size,
+                            const scalar_t* bias,
+                            scalar_t* out_bias,
+                            int64_t* alias) {
+  scalar_t eps = 1e-6;
+
+  // Calculate the average bias
+  for (int64_t i = 0; i < rowptr_size - 1; i++) {
```
yaoyaowd, line 231: parallel for?

pyg_lib/csrc/random/cpu/biased_sampling.h
```diff
@@ -127,8 +127,166 @@ index_t biased_random_alias(const index_t* idx,
...
+    int64_t* alias_beg = alias + rowptr_data[i];
+    scalar_t avg = 0;
+
+    for (int64_t j = 0; j < len; j++) {
+      avg += beg[j];
+    }
+    avg /= len;
+
+    // The sets for index with a bias lower or higher than average
+    std::vector<std::pair<int64_t, scalar_t>> high, low;
```
yaoyaowd, line 244: performance wise, we can reserve space for high and low.

pyg_lib/csrc/random/cpu/biased_sampling.h
```diff
@@ -127,8 +127,166 @@ index_t biased_random_alias(const index_t* idx,
...
+      low.pop_back();
+      high.pop_back();
+
+      // Handle the lower one:
+      out_beg[low_idx] = low_bias / avg;
+      alias_beg[low_idx] = high_idx;
+
+      // Handle the higher one:
+      scalar_t high_bias_left = high_bias - (avg - low_bias);
+      out_beg[high_idx] = 1;
```
yaoyaowd, line 277: I feel that we should not do this unless high_bias_left is close to avg, i.e else condition after line 285.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.12)
 project(pyg)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
rusty1s, line 3: Any reason for this? I would like to keep this the same as the one enforced by PyTorch.

pyg_lib/csrc/random/cpu/biased_sampling.cpp
```diff
@@ -0,0 +1,60 @@
+#include "pyg_lib/csrc/random/cpu/biased_sampling.h"
+
+namespace pyg {
+namespace random {
+
+at::Tensor biased_to_cdf(at::Tensor rowptr, at::Tensor bias) {
```
rusty1s, line 6:
```suggestion
at::Tensor biased_to_cdf(const at::Tensor& rowptr, const at::Tensor& bias) {
```

pyg_lib/csrc/random/cpu/biased_sampling.cpp
```diff
@@ -0,0 +1,60 @@
...
+  TORCH_CHECK(rowptr.is_cpu(), "'rowptr' must be a CPU tensor");
+  TORCH_CHECK(bias.is_cpu(), "'bias' must be a CPU tensor");
+
+  auto cdf = at::empty_like(bias);
```
rusty1s, line 10:
```suggestion
  const auto cdf = at::empty_like(bias);
```

pyg_lib/csrc/random/cpu/biased_sampling.cpp
```diff
@@ -0,0 +1,60 @@
...
+  int64_t rowptr_size = rowptr.size(0);
+  int64_t* rowptr_data = rowptr.data_ptr<int64_t>();
```
rusty1s, line 12: I would like to keep `rowptr` data-types dynamic. How about we wrap this inside a `AT_DISPATCH_INTEGRAL_TYPES` call?

pyg_lib/csrc/random/cpu/biased_sampling.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  AT_DISPATCH_FLOATING_TYPES(bias.scalar_type(), "biased_to_cdf", [&] {
+    scalar_t* bias_data = bias.data_ptr<scalar_t>();
```
rusty1s, line 15:
```suggestion
    auto bias_data = bias.data_ptr<scalar_t>();
```

test/csrc/random/test_biased_random.cpp
```diff
@@ -74,3 +75,50 @@ TEST(BiasedSamplingAliasRandomTest, BasicAssertions) {
...
+TEST(BiasedSamplingCDFConversionTest, BasicAssertions) {
+  auto options = at::TensorOptions().dtype(at::kLong);
+
+  auto graph = cycle_graph(/*num_nodes=*/4, options);
+  auto rowptr = std::get<0>(graph);
+
+  std::vector<float> bias_vec{1.5, 0.5, 0.8, 0.2, 0.1, 0.3, 1.0, 1.0};
+  std::vector<float> cdf_vec{0.0, 0.75, 0.0, 0.8, 0.0, 0.25, 0.0, 0.5};
+
+  at::Tensor bias = at::from_blob(bias_vec.data(), {bias_vec.size()},
```
rusty1s, line 88: Use `from_vector` in `csrc/cpu/utils.h` for this.

CMakeLists.txt
ZenoTan: It's mainly for the convenience of tuple unpacking and `std::optional`. I could change them back by using `c10::optional` and manually unpack tuples.

pyg_lib/csrc/random/cpu/biased_sampling.cpp
ZenoTan: We need to write data to cdf so it's not a const here

pyg_lib/csrc/random/cpu/biased_sampling.cpp
```diff
@@ -0,0 +1,166 @@
...
+          int64_t* alias_beg = alias + rowptr_data[i];
+          scalar_t avg = 0;
+
+          for (int64_t j = 0; j < len; j++) {
+            avg += beg[j];
+          }
+          avg /= len;
+
+          // The sets for index with a bias lower or higher than average
+          std::vector<std::pair<int64_t, scalar_t>> high, low;
```
yaoyaowd, line 110: One more optimization we can do is to use `std::vector<size_t> high, low` instead of pair.
Later you can do
```
scalar_t low_bias = out_beg[low_idx];
out_beg[low_idx] = low_bias / avg;
out_beg[high_idx] -= (avg - low_bias);
```
Create pairs are much more costly than just use int.

yaoyaowd (commented):
One more nit comment is to always use `size_t` for indexes such as
```
for (size_t i = _s; i < _e; ++i)
for (size_t j = 0; j < len; ++j)
```
Both from complier optimization and readability point of view.

yaoyaowd (approved):

pyg_lib/csrc/random/cpu/biased_sampling.cpp
rusty1s: The tensor will always be a const since we only change its data, not its metadata. It‘s not so important though :)

PR #42, opened 2022-05-03
`pyg::subgraph` CUDA implementation

rusty1s:


pyg_lib/csrc/sampler/cuda/subgraph_kernel.cu
```diff
@@ -8,6 +8,8 @@ namespace sampler {
 
 namespace {
 
+#define FULL_MASK 0xffffffff
+
```
ZenoTan, line 12: Could we define `WARP` 32 here?

ZenoTan: Better if in a `utils` header

pyg_lib/csrc/sampler/cuda/random_walk_kernel.cu
```diff
@@ -1,29 +1,13 @@
...
-  const auto props = at::cuda::getCurrentDeviceProperties();
-  const auto blocks_per_sm = props->maxThreadsPerMultiProcessor / 256;
-  const auto max_blocks = props->multiProcessorCount * blocks_per_sm;
-  const auto max_threads = threads();
-  return std::min(max_blocks, (numel + max_threads - 1) / max_threads);
-}
-
-#define CUDA_1D_KERNEL_LOOP(scalar_t, i, n)                           \
-  for (scalar_t i = (blockIdx.x * blockDim.x) + threadIdx.x; i < (n); \
-       i += (blockDim.x * gridDim.x))
```
yaoyaowd, line 25: I believe we assume gridDim.x is 1 by default. If not, the loop will be
```
for (scalar_t i = (blockIdx.x * blockDim.x * gridDim.x) + threadIdx.x; i < (n); ...
```

pyg_lib/csrc/sampler/cuda/subgraph_kernel.cu
```diff
@@ -0,0 +1,90 @@
...
+  c10::optional<at::Tensor> out_edge_id = c10::nullopt;
+
+  AT_DISPATCH_INTEGRAL_TYPES(nodes.scalar_type(), "subgraph_kernel", [&] {
+    const auto rowptr_data = rowptr.data_ptr<scalar_t>();
+    const auto col_data = col.data_ptr<scalar_t>();
+    const auto nodes_data = nodes.data_ptr<scalar_t>();
+    const auto to_local_node_data = to_local_node.data_ptr<scalar_t>();
+    auto deg_data = deg.data_ptr<scalar_t>();
+
+    // Compute induced subgraph degree, parallelize with 32 threads per node:
```
yaoyaowd, line 70: I'm actually not sure if it is necessary to parallelize with 32 threads per nodes. Most of the time we are dealing with sparse data and a lot of threads will not go into for loop.

If you are looking for extreme performance, you can bundle `to_local_node_data` and `col_data` into one iterator structure and use this function. I haven't seen any better performance than it in the past.
`https://nvlabs.github.io/cub/structcub_1_1_device_segmented_reduce.html#a4854a13561cb66d46aa617aab16b8825`

pyg_lib/csrc/sampler/cuda/subgraph_kernel.cu
```diff
@@ -0,0 +1,90 @@
...
+    const bool return_edge_id) {
+  TORCH_CHECK(rowptr.is_cuda(), "'rowptr' must be a CUDA tensor");
+  TORCH_CHECK(col.is_cuda(), "'col' must be a CUDA tensor");
+  TORCH_CHECK(nodes.is_cuda(), "'nodes' must be a CUDA tensor");
+
+  const auto stream = at::cuda::getCurrentCUDAStream();
+
+  // We maintain a O(N) vector to map global node indices to local ones.
+  // TODO Can we do this without O(N) storage requirement?
+  const auto to_local_node = nodes.new_full({rowptr.size(0) - 1}, -1);
```
ZenoTan, line 54: Does N means the number of nodes in the graph?
What if we could filter each node in `nodes_data` since it should be much smaller than `rowptr_data`.
Otherwise we may consider caching this tensor to reduce memory allocation for each time.

rusty1s: Good points! We use this vector as the inverse mapping from global node indices to new local ones. In C++, we use a map for this but can't do the same in CUDA. I will ping NVIDIA if they have an elegant solution for this.

Caching is an option as well, but requires a (non-intuitive and backend-specific) change in input arguments. I added it as a TODO for now.

pyg_lib/csrc/sampler/cuda/subgraph_kernel.cu
rusty1s: Do you have an example of bundling `to_local_node_data` and `col_data` into one iterator structure? This looks really interesting.

I am okay with dropping the warp-level parallelism for now, but we will lose the contiguous access to `col_data`, and probably under-utilize the number of threads available on modern GPUs.

rusty1s: On a second look, this doesn't seem possible since `col_data` refers to edges, while `to_local_node_data` refers to nodes, while we actually want do the compute across the number of nodes in the induced subgraph.

pyg_lib/csrc/sampler/cuda/subgraph_kernel.cu
ZenoTan: There's GPU hash table/set which may require some atomic operations when you build it, but lookup is fast.
I found that caching is not a good option since you have to reset the array every time.

Since you can sample on GPU, then the graph is not that big, a node array is not that bad and can make the code less complicated

yaoyaowd (approved):
Let's see if we can improve it later.

PR #43, opened 2022-05-04
Hetero subgraph with dispatching

ZenoTan:


yaoyaowd (dismissed):

pyg_lib/csrc/sampler/cpu/mapper.h
```diff
@@ -10,6 +10,8 @@ namespace sampler {
 template <typename scalar_t>
 class Mapper {
  public:
+  using type = scalar_t;
```
yaoyaowd, line 13: why need this using?

pyg_lib/csrc/sampler/cpu/mapper.h
```diff
@@ -23,26 +25,28 @@ class Mapper {
 
   void fill(const scalar_t* nodes_data, const scalar_t size) {
     if (use_vec) {
-      for (scalar_t i = 0; i < size; ++i)
+      for (scalar_t i = 0; i < size; ++i) {
```
yaoyaowd, line 28: Let me post my question here, I read some documents and based on my understanding scalar_t includes both float, double, int32, int64 during compile. But in a lot of our usecases we are iterating over integers. How does pytorch avoid compile float type for these functions? Is there a better way to be more specific to the data types here?

pyg_lib/csrc/utils/types.h
```diff
@@ -0,0 +1,34 @@
...
+#include <string>
+
+#include <ATen/ATen.h>
+
+namespace pyg {
+namespace utils {
+
+const std::string SPLIT_TOKEN = "__";
+
+using edge_t = std::string;
```
yaoyaowd, line 12: If we stick with Google C++ Style Guide, the type name will be upper case
```
using EdgeType = std::string;
...
using EdgeTensorDict = c10::Dict<EdgeType, at::Tensor>;
```

pyg_lib/csrc/sampler/cpu/subgraph_kernel.cpp
```diff
@@ -15,78 +16,7 @@ std::tuple<at::Tensor, at::Tensor, c10::optional<at::Tensor>> subgraph_kernel(
...
-              out_edge_id_data[offset] = j;
-            offset++;
-          }
-        }
-      }
-    });
-  });
-
-  return std::make_tuple(out_rowptr, out_col, out_edge_id);
+  return subgraph_bipartite(rowptr, col, nodes, nodes, return_edge_id);
```
yaoyaowd, line 19: The code structure looks a little weird to me because `csrc/sampler/cpu/subgraph_kernel` exists for register `TORCH_LIBRARY_IMPL` and it is using a general implementation in `csr/sampler/subgraph.cpp`. How about reorganize the code like this:
```
csr
  - ops
    # all ops expose for pytorch.
    - sampler
  # all general graph operation.
  - sampler
```
We don't need to refactor the code structure now. But want to hear your opinion.

yaoyaowd: nvm, seems subgraph.cpp also defines library. Why not merge them together since sampler/subgraph.cpp also runs on cpu only.

pyg_lib/csrc/utils/hetero_dispatch.h
```diff
@@ -0,0 +1,191 @@
...
+      : rowptr_(rowptr), col_(col), op_(op) {
+    // Check early
+    static_assert(is_std_function<T>::value, "Must pass a function");
+  }
+
+  template <typename... Args>
+  c10::Dict<edge_t, result_type> operator()(Args&&... args) {
+    c10::Dict<edge_t, result_type> dict;
+    for (const auto& kv : rowptr_) {
+      auto edge = kv.key();
```
yaoyaowd, line 171: I think after you rename the type, the code will be either to read. I.e `edge_t` become `EdgeType`, `result_type` become `ResultType` since they are types.
Here 
```
auto edge_type = kv.key();
```
So that we know it is a type instead of a tensor.

pyg_lib/csrc/sampler/cpu/mapper.h
ZenoTan: It's previously to deduce the element type, but now it's not used and I will remove it

pyg_lib/csrc/sampler/cpu/mapper.h
ZenoTan: There are some helper functions like `is_integral` for a `dtype`, but IMO it is mostly runtime checking. We can also use some STL type checking for compile time.

pyg_lib/csrc/sampler/cpu/subgraph_kernel.cpp
ZenoTan: We could follow the style in other pyg repos: put CPU/GPU specific impl in separate folders and provide common interface in a higher directory.

pyg_lib/csrc/sampler/cpu/subgraph_kernel.cpp
```diff
@@ -3,96 +3,139 @@
...
+  AT_DISPATCH_INTEGRAL_TYPES(
+      src_nodes.scalar_type(), "subgraph_bipartite_kernel", [&] {
+        // TODO: at::max parallel but still a little expensive
+        Mapper<scalar_t> mapper(at::max(col).item<scalar_t>() + 1,
+                                dst_nodes.size(0));
+        mapper.fill(dst_nodes);
+
+        auto res = subgraph_with_mapper<scalar_t>(rowptr, col, src_nodes,
+                                                  mapper, return_edge_id);
+        out_rowptr = std::get<0>(res);
```
yaoyaowd, line 117: or maybe we could do `std::tie(out_powptr, out_col, out_edge_id) = res`?

pyg_lib/csrc/sampler/subgraph.cpp
```diff
@@ -25,10 +28,76 @@ std::tuple<at::Tensor, at::Tensor, c10::optional<at::Tensor>> subgraph(
...
+  utils::HeteroDispatchArg<utils::NodeTensorDict, at::Tensor,
+                           utils::NodeDstMode>
+      dst_nodes_arg(dst_nodes);
+  utils::HeteroDispatchArg<c10::Dict<utils::EdgeType, bool>, bool,
+                           utils::EdgeMode>
+      edge_id_arg(return_edge_id);
+
+  for (const auto& kv : rowptr) {
+    const auto& edge_type = kv.key();
+    bool pass = filter_args_by_edge(edge_type, src_nodes_arg, dst_nodes_arg,
```
yaoyaowd, line 76: I'd still prefer
```
pass = src_nodes_args.filter_by_edge(edge_type) && dst_nodes_args.filter_by_edge(edge_type) && edge_id_arg.filter_by_edge(edge_type)
```
or from an efficiency point of view.
```
auto dst = get_dst(edge_type)
auto src = get_src(edge_type)
bool pass = return_edge_id.counts(edge_type) > 0 && src_nodes.counts(src) > 0 && dst_nodes.counts(dst) > 0;
```

pyg_lib/csrc/sampler/subgraph.cpp
```diff
@@ -25,10 +28,76 @@ std::tuple<at::Tensor, at::Tensor, c10::optional<at::Tensor>> subgraph(
...
+                                    edge_id_arg);
+    if (pass) {
+      auto vals = value_args_by_edge(edge_type, src_nodes_arg, dst_nodes_arg,
+                                     edge_id_arg);
+      const auto& r = rowptr.at(edge_type);
+      const auto& c = col.at(edge_type);
+      res.insert(edge_type,
+                 subgraph_bipartite(r, c, std::get<0>(vals), std::get<1>(vals),
```
yaoyaowd, line 84: and here would just be
```
subgraph_bipartite(r, c, src_nodes.at(src), dst_nodes.at(dst), return_edge_id.at(edge_type));
```

pyg_lib/csrc/sampler/cpu/mapper.h
rusty1s: The `AT_DISPATCH_INTEGRAL_TYPES` call handles which types `scalar_t` can take (during compile time).

pyg_lib/csrc/sampler/cpu/subgraph_kernel.cpp
```diff
@@ -3,96 +3,139 @@
...
-        for (scalar_t j = rowptr_data[v]; j < rowptr_data[v + 1]; ++j) {
-          const auto w = mapper.map(col_data[j]);
-          if (w >= 0) {
-            out_col_data[offset] = w;
-            if (return_edge_id)
-              out_edge_id_data[offset] = j;
-            offset++;
-          }
+  AT_DISPATCH_INTEGRAL_TYPES(
+      nodes.scalar_type(), "subgraph_kernel_with_mapper", [&] {
```
rusty1s, line 27: Can we make this a one-liner again?

pyg_lib/csrc/sampler/cpu/subgraph_kernel.cpp
```diff
@@ -3,96 +3,139 @@
...
+  TORCH_CHECK(src_nodes.is_cpu(), "'src_nodes' must be a CPU tensor");
+  TORCH_CHECK(dst_nodes.is_cpu(), "'dst_nodes' must be a CPU tensor");
+
+  const auto num_nodes = rowptr.size(0) - 1;
+  at::Tensor out_rowptr, out_col;
+  c10::optional<at::Tensor> out_edge_id;
+
+  AT_DISPATCH_INTEGRAL_TYPES(
+      src_nodes.scalar_type(), "subgraph_bipartite_kernel", [&] {
+        // TODO: at::max parallel but still a little expensive
```
rusty1s, line 110: Let's add a `size` argument here. The `max` should definitely be avoided.

pyg_lib/csrc/sampler/cpu/subgraph_kernel.cpp
rusty1s: +1

pyg_lib/csrc/sampler/cpu/subgraph_kernel.cpp
```diff
@@ -3,96 +3,139 @@
...
+    const at::Tensor& nodes,
+    const bool return_edge_id) {
+  return subgraph_bipartite_kernel(rowptr, col, nodes, nodes, return_edge_id);
+}
+
 }  // namespace
 
 TORCH_LIBRARY_IMPL(pyg, CPU, m) {
   m.impl(TORCH_SELECTIVE_NAME("pyg::subgraph"), TORCH_FN(subgraph_kernel));
+  m.impl(TORCH_SELECTIVE_NAME("pyg::subgraph_bipartite"),
```
rusty1s, line 137: Any reason we want to expose that? Looks more like an internal function to me.

pyg_lib/csrc/sampler/subgraph.cpp
```diff
@@ -25,10 +28,76 @@ std::tuple<at::Tensor, at::Tensor, c10::optional<at::Tensor>> subgraph(
...
+  at::checkAllSameType(c, {rowptr_t, col_t, src_nodes_t, dst_nodes_t});
+
+  static auto op = c10::Dispatcher::singleton()
+                       .findSchemaOrThrow("pyg::subgraph_bipartite", "")
+                       .typed<decltype(subgraph_bipartite)>();
+  return op.call(rowptr, col, src_nodes, dst_nodes, return_edge_id);
+}
+
+c10::Dict<utils::EdgeType,
+          std::tuple<at::Tensor, at::Tensor, c10::optional<at::Tensor>>>
```
rusty1s, line 53: IMO, the output should be a tuple of dictionaries (similar to the input).

pyg_lib/csrc/sampler/subgraph.cpp
```diff
@@ -25,10 +28,76 @@ std::tuple<at::Tensor, at::Tensor, c10::optional<at::Tensor>> subgraph(
...
+
+  for (const auto& kv : rowptr) {
+    const auto& edge_type = kv.key();
+    bool pass = src_nodes_arg.filter_by_edge(edge_type) &&
+                dst_nodes_arg.filter_by_edge(edge_type) &&
+                edge_id_arg.filter_by_edge(edge_type);
+    if (pass) {
+      const auto& r = rowptr.at(edge_type);
+      const auto& c = col.at(edge_type);
+      res.insert(edge_type, subgraph_bipartite(
```
rusty1s, line 82: Shouldn't we user the mapper here? Other-wise, we will re-map across every edge type.

pyg_lib/csrc/utils/types.h
```diff
@@ -0,0 +1,34 @@
...
+
+inline RelationType get_rel(const EdgeType& e) {
+  auto beg = e.find_first_of(SPLIT_TOKEN) + SPLIT_TOKEN.size();
+  return e.substr(beg,
+                  e.find_last_of(SPLIT_TOKEN) - SPLIT_TOKEN.size() + 1 - beg);
+}
+
+inline NodeType get_dst(const EdgeType& e) {
+  return e.substr(e.find_last_of(SPLIT_TOKEN) + 1);
+}
```
rusty1s, line 31: We could also add a function that to map tuples to strings and vice versa.

pyg_lib/csrc/sampler/subgraph.cpp
```diff
@@ -25,10 +28,76 @@ std::tuple<at::Tensor, at::Tensor, c10::optional<at::Tensor>> subgraph(
...
+                       .findSchemaOrThrow("pyg::subgraph_bipartite", "")
+                       .typed<decltype(subgraph_bipartite)>();
+  return op.call(rowptr, col, src_nodes, dst_nodes, return_edge_id);
+}
+
+c10::Dict<utils::EdgeType,
+          std::tuple<at::Tensor, at::Tensor, c10::optional<at::Tensor>>>
+hetero_subgraph(const utils::EdgeTensorDict& rowptr,
+                const utils::EdgeTensorDict& col,
+                const utils::NodeTensorDict& src_nodes,
```
rusty1s, line 56: Not sure why we have both `src_nodes` and `dst_nodes`. IMO, these can be safely merged as in https://pytorch-geometric.readthedocs.io/en/latest/modules/data.html#torch_geometric.data.HeteroData.subgraph.

pyg_lib/csrc/sampler/subgraph.cpp
ZenoTan: Yes it has a cost, but the mapper is more read-intensive. I will add a TODO here.

pyg_lib/csrc/utils/types.h
ZenoTan: Good idea.

pyg_lib/csrc/sampler/subgraph.cpp
ZenoTan: Separating `src` and `dst` is just to give some flexibility. We could also have the merged API though.

pyg_lib/csrc/sampler/cpu/subgraph_kernel.cpp
ZenoTan: If the user want to build a subgraph of a bipartite graph then he can use it.

PR #61, opened 2022-07-07
Autograd for `segment_matmul`

puririshi98:


pyg_lib/csrc/ops/matmul.cpp
```diff
@@ -1,33 +1,80 @@
...
 namespace ops {
 
+using torch::autograd::AutogradContext;
+using torch::autograd::Variable;
+using torch::autograd::variable_list;
+
 // Performs matrix multiplication across list of elements.
-std::vector<at::Tensor> grouped_matmul(const std::vector<at::Tensor>& input,
-                                       const std::vector<at::Tensor>& other) {
-  // TODO (matthias) Add TensorArg definitions.
```
rusty1s, line 12: Can we keep the TODOs?

pyg_lib/csrc/ops/matmul.cpp
```diff
@@ -1,33 +1,80 @@
...
+    auto other_grad = op.call(grad_outs, other);
+    return {input_grad, other_grad};
+  }
+};
 
 // Performs matrix multiplication according to segments.
-at::Tensor segment_matmul(const at::Tensor& input,
-                          const at::Tensor& ptr,
-                          const at::Tensor& other) {
-  // TODO (matthias) Add TensorArg definitions.
```
rusty1s, line 25: Again.

rusty1s (approved):
I think this perfect. A test would be great to ensure correctness.

pyg_lib/csrc/ops/matmul.cpp
```diff
@@ -1,33 +1,80 @@
...
+  static variable_list backward(AutogradContext* ctx, variable_list grad_outs) {
+    auto saved = ctx->get_saved_variables();
+    auto out = saved[0];
+    auto input = saved[1];
+    for (size_t i = 0; i < input.size(); ++i)
+      input[i] = input[i].transpose(-2, -1);
+    auto other = saved[2];
+    for (size_t i = 0; i < input.size(); ++i)
+      other[i] = other[i].transpose(-2, -1);
+    auto input_grad = op.call(input, grad_outs);
```
rusty1s, line 36: I think it is pretty important that we only compute these if `input` requires grad.

pyg_lib/csrc/ops/matmul.cpp
```diff
@@ -1,33 +1,80 @@
...
+    auto other_grad = op.call(grad_outs, other);
```
rusty1s, line 37: Here again.

puririshi98:
I made some good progress on this PR but there are a few remaining errors that the compiler complains about. Not sure on the best way to deal with each issue so I thought itd be good to discuss.
- this line in grouped matmul: [ctx->save_for_backward({input, other});](https://github.com/pyg-team/pyg-lib/blob/4ebfe1f4d2a1820ecf9df00b6f959012c452f512/pyg_lib/csrc/ops/matmul.cpp#L29)
    - in this case input and other are of type variable_list but save_for_backward expects a variable list as input and so cant take in the list of variable_list i am passing. there are two options here:
        - if u know of an alternative to save_for_backward which has the same functionality but allows for more general input types
        - we could concatenate the two, but then i would also have to store the length of each to separate them back out in the backward pass
- for the [grouped_matmul backward pass](https://github.com/pyg-team/pyg-lib/blob/4ebfe1f4d2a1820ecf9df00b6f959012c452f512/pyg_lib/csrc/ops/matmul.cpp#L33-L44), in the case that input does require grad, we are returning 2 variable lists. this brings up a similar issue to the last one.
    - we can concatenate them, although I am not sure how autograd expects the outputs to be organized
    - we can keep it as is and change the return type of the function to be an std vector of variable_list type, but i am unsure how autograd handles this
- lastly for the [segment_matmul backward pass](https://github.com/pyg-team/pyg-lib/blob/4ebfe1f4d2a1820ecf9df00b6f959012c452f512/pyg_lib/csrc/ops/matmul.cpp#L64-L76) in the normal case we return a single variable, however in the case that input also requires gradient to be computed, we are returning 2 variables as a variable list, which the compiler complains about since the returned type no longer matches the expected return type of the function. not sure what to do in this case as the return type depends on the conditional.

rusty1s:
@puririshi98 I am okay for now to only implement backward in `segment_matmul`. For your problem, I think you always need to return a list of variables (where some of them can be empty, see [here](https://github.com/rusty1s/pytorch_scatter/blob/master/csrc/scatter.cpp#L106)).

puririshi98:
```
2: Test command: /home/runner/work/pyg-lib/pyg-lib/build/test_biased_random "--gtest_filter=BiasedSamplingAliasRandomTest.BasicAssertions" "--gtest_also_run_disabled_tests"
2: Test timeout computed to be: 10000000
2: Running main() from /home/runner/work/pyg-lib/pyg-lib/build/_deps/googletest-src/googletest/src/gtest_main.cc
2: Note: Google Test filter = BiasedSamplingAliasRandomTest.BasicAssertions
2: [==========] Running 1 test from 1 test suite.
2: [----------] Global test environment set-up.
2: [----------] 1 test from BiasedSamplingAliasRandomTest
2: [ RUN      ] BiasedSamplingAliasRandomTest.BasicAssertions
2: /home/runner/work/pyg-lib/pyg-lib/test/csrc/random/test_biased_random.cpp:74: Failure
2: Expected: (cnt[1]) < (0.7 * iter), actual: 701 vs 700
2: [  FAILED  ] BiasedSamplingAliasRandomTest.BasicAssertions (1 ms)
2: [----------] 1 test from BiasedSamplingAliasRandomTest (1 ms total)
2: 
2: [----------] Global test environment tear-down
2: [==========] 1 test from 1 test suite ran. (1 ms total)
2: [  PASSED  ] 0 tests.
2: [  FAILED  ] 1 test, listed below:
2: [  FAILED  ] BiasedSamplingAliasRandomTest.BasicAssertions
2: 
2:  1 FAILED TEST
```
the last CI failure seems unrelated to my changes.

puririshi98:
also unsure how to integrate the python test I wrote as part of this PR into the CI for this project

PR #63, opened 2022-07-13
backward pass for other in segment_matmul

puririshi98:


puririshi98:
subsumed by https://github.com/pyg-team/pyg-lib/pull/64

PR #73, opened 2022-08-03
Speed-up CUTLASS implementation

puririshi98:


rusty1s:
Thanks @puririshi98. Won't we need to reshape the output back to the original shape? If yes, let's take this in a separate PR.

PR #79, opened 2022-08-31
Add neighbor sampler benchmark

kgajdamo:
- Added neighbor sampler benchmark to pyg-lib.
- Benchmark measures performance for the neighbor_sample from pyg-lib as well as neighbor_sample from pytorch_sparse.
- Added argument "return_csc" to withDataset decorator to optionally return matrix in CSC format (for the purpose of pytorch_sparse neighbor sample).

### DIFF
--- a/benchmark/sampler/neighbor.py
+++ b/benchmark/sampler/neighbor.py
@@ -1,0 +1,62 @@
+import argparse
+import time
+
+import torch
+# uncomment below to enable torch.ops.torch_sparse.neighbor_sample
+# import torch_sparse
+from tqdm import tqdm
+
+import pyg_lib
+from pyg_lib.testing import withDataset, withSeed
+
+argparser = argparse.ArgumentParser('Neighbor Sampler benchmark')
+argparser.add_argument('--batch-sizes', nargs='+',
+                       default=[512, 1024, 2048, 4096, 8192], type=int)
+argparser.add_argument('--num_neighbors', default=[[-1], [15, 10, 5],
+                                                   [20, 15, 10]], type=int)
+argparser.add_argument('--replace', default=False, type=bool)
+argparser.add_argument('--directed', default=True, type=bool)
+
+args = argparser.parse_args()
+
+
+@withSeed
+@withDataset('DIMACS10', 'citationCiteseer', True)
+def test_neighbor(dataset, **kwargs):
+    (rowptr, col, colptr, row), num_nodes = dataset, dataset[0].size(0) - 1
+
+    for num_neighbors in args.num_neighbors:
+        for batch_size in args.batch_sizes:
+            # pyg-lib neighbor sampler
+            start = time.perf_counter()
+            for node in tqdm(range(num_nodes - batch_size)):
+                seed = torch.arange(node, node + batch_size)
+                pyg_lib.sampler.neighbor_sample(rowptr, col, seed,
+                                                num_neighbors, args.replace,
+                                                args.directed, disjoint=False,
+                                                return_edge_id=True)
+            stop = time.perf_counter()
+            print('-------------------------')
+            print('pyg-lib neighbor sample')
+            print(f'Batch size={batch_size}, '
+                  f'Num_neighbors={num_neighbors}, '
+                  f'Inference time={stop-start:.3f} seconds\n')
+
+            # pytorch-sparse neighbor sampler
+            start = time.perf_counter()
+            for node in tqdm(range(num_nodes - batch_size)):
+                seed = torch.arange(node, node + batch_size)
+                torch.ops.torch_sparse.neighbor_sample(colptr, row, seed,
+                                                       num_neighbors,
+                                                       args.replace,
+                                                       args.directed)
+            stop = time.perf_counter()
+            print('-------------------------')
+            print('pytorch_sparse neighbor sample')
+            print(f'Batch size={batch_size}, '
+                  f'Num_neighbors={num_neighbors}, '
+                  f'Inference time={stop-start:.3f} seconds\n')
+
+
+if __name__ == '__main__':
+    test_neighbor()
--- a/benchmark/sampler/neighbor.py
+++ b/benchmark/sampler/neighbor.py
@@ -2,8 +2,12 @@ import argparse
 import time
 
 import torch
-# uncomment below to enable torch.ops.torch_sparse.neighbor_sample
-# import torch_sparse
+
+try:
+    import torch_sparse  # noqa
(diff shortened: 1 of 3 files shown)


rusty1s:
Cool! How do the two implementations compare to each other?

benchmark/sampler/neighbor.py
```diff
@@ -0,0 +1,62 @@
+import argparse
+import time
+
+import torch
+# uncomment below to enable torch.ops.torch_sparse.neighbor_sample
+# import torch_sparse
```
mananshah99, line 6: Can we gate this with a `try/catch`, and avoid the `torch_sparse` benchmark in case it was not successfully imported?

benchmark/sampler/neighbor.py
```diff
@@ -0,0 +1,62 @@
...
+argparser.add_argument('--num_neighbors', default=[[-1], [15, 10, 5],
+                                                   [20, 15, 10]], type=int)
+argparser.add_argument('--replace', default=False, type=bool)
+argparser.add_argument('--directed', default=True, type=bool)
+
+args = argparser.parse_args()
+
+
+@withSeed
+@withDataset('DIMACS10', 'citationCiteseer', True)
```
mananshah99, line 24:
```suggestion
@withDataset('DIMACS10', 'citationCiteseer', return_csc=True)
```
nit, since I was confused by L26

benchmark/sampler/neighbor.py
```diff
@@ -0,0 +1,62 @@
...
+def test_neighbor(dataset, **kwargs):
+    (rowptr, col, colptr, row), num_nodes = dataset, dataset[0].size(0) - 1
+
+    for num_neighbors in args.num_neighbors:
+        for batch_size in args.batch_sizes:
+            # pyg-lib neighbor sampler
+            start = time.perf_counter()
+            for node in tqdm(range(num_nodes - batch_size)):
```
mananshah99, line 32: Do we instead want
```
range(0, num_nodes, batch_size)
```

benchmark/sampler/neighbor.py
```diff
@@ -0,0 +1,62 @@
...
+            stop = time.perf_counter()
+            print('-------------------------')
+            print('pyg-lib neighbor sample')
+            print(f'Batch size={batch_size}, '
+                  f'Num_neighbors={num_neighbors}, '
+                  f'Inference time={stop-start:.3f} seconds\n')
+
+            # pytorch-sparse neighbor sampler
+            start = time.perf_counter()
+            for node in tqdm(range(num_nodes - batch_size)):
```
mananshah99, line 47: Same question

pyg_lib/testing.py
```diff
@@ -52,9 +54,12 @@ def get_sparse_matrix(
     name: str,
     dtype: torch.dtype = torch.long,
     device: Optional[torch.device] = None,
-) -> Tuple[Tensor, Tensor]:
+    return_csc: Optional[bool] = False,
+) -> Tuple[Tensor, Tensor, Optional[Tensor], Optional[Tensor]]:
```
mananshah99, line 58: This return type is a little confusing to me. Perhaps we can provide a separate utility function that handles CSC conversion?

kgajdamo:
> Cool! How do the two implementations compare to each other?

@rusty1s Just like it is done in a [subgraph benchmark](https://github.com/pyg-team/pyg-lib/blob/master/benchmark/main.py), we print out the total duration of the neighbor_sample function first for pyg-lib and then for the pytorch_sparse.

pyg_lib/testing.py
```diff
@@ -27,14 +27,16 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def withDataset(group: str, name: str) -> Callable:
+def withDataset(group: str, name: str,
+                return_csc: Optional[bool] = False) -> Callable:
```
rusty1s, line 31: I don't think we need this option. We can call both samplers just with the returned CSR type. It will result in the same computation.

benchmark/sampler/neighbor.py
rusty1s: +1, something like
```python
try:
    import torch_sparse
    baseline_neighbor_sample = torch.ops.torch_sparse.neighbor_sample
except ImportError:
    baseline_neighbor_sample = None
```

benchmark/sampler/neighbor.py
```diff
@@ -0,0 +1,62 @@
...
+
+import pyg_lib
+from pyg_lib.testing import withDataset, withSeed
+
+argparser = argparse.ArgumentParser('Neighbor Sampler benchmark')
+argparser.add_argument('--batch-sizes', nargs='+',
+                       default=[512, 1024, 2048, 4096, 8192], type=int)
+argparser.add_argument('--num_neighbors', default=[[-1], [15, 10, 5],
+                                                   [20, 15, 10]], type=int)
+argparser.add_argument('--replace', default=False, type=bool)
```
rusty1s, line 17: Use `action="store_true"` instead.

pyg_lib/testing.py
kgajdamo: Good idea, thanks.

benchmark/sampler/neighbor.py
kgajdamo: Thanks for the comments! The problem with this import is that basically torch_sparse is not used in this file and we get an error from pre-commit:
`benchmark/sampler/neighbor.py:7:5: F401 'torch_sparse' imported but unused`
So there are two options: we can comment out this line or create a dummy torch_sparse object. What do You think?

rusty1s: Ah, simply do
```python
import torch_sparse  # noqa
```

kgajdamo: Thanks, done.

benchmark/sampler/neighbor.py
kgajdamo: Done.

pyg_lib/testing.py
kgajdamo: Done.

benchmark/sampler/neighbor.py
kgajdamo: Done.

pyg_lib/testing.py
kgajdamo: According to the [Matthias comment](https://github.com/pyg-team/pyg-lib/pull/79#discussion_r960343916) it has been removed.

rusty1s (approved):

rusty1s:
Thank you!

PR #84, opened 2022-09-05
Add `IndexTracker` as faster alternative to `unordered_set`

mszarma:
Apply usage of indicesTracker mechanism in neighbor sample.
IndicesTracker is replacing the unordered_set usage by using vector<char> indexing or flat_hash_set (base on population size).
Contributors/Kudos: @kgajdamo @OlhaBabicheva 
image
image

### DIFF
--- a/pyg_lib/csrc/sampler/cpu/indices_tracker.h
+++ b/pyg_lib/csrc/sampler/cpu/indices_tracker.h
@@ -1,0 +1,49 @@
+#pragma once
+
+#include <ATen/ATen.h>
+
+#include "parallel_hashmap/phmap.h"
+
+namespace pyg {
+namespace sampler {
+
+template <typename scalar_t>
+class IndicesTracker {
+ public:
+  IndicesTracker(const scalar_t& size) : size(size) {
+    // TODO: add better switching threshold value mechanism?
+    use_vec = (size < 100000);
+    if (use_vec)
+      rnd_indices_vec.resize(size, 0);
+  }
+
+  bool tryInsert(const scalar_t& index) {
+    if (use_vec) {
+      if (rnd_indices_vec[index] == 0) {
+        rnd_indices_vec[index] = 1;
+        return true;
+      } else {
+        return false;
+      }
+    } else {
+      return rnd_indices_set.insert(index).second;
+    }
+  }
+
+  void insert(const scalar_t& index) {
+    if (use_vec) {
+      rnd_indices_vec[index] = 1;
+    } else {
+      rnd_indices_set.insert(index);
+    }
+  }
+
+ private:
+  const scalar_t& size;
+  bool use_vec;
+  std::vector<char> rnd_indices_vec;
+  phmap::flat_hash_set<scalar_t> rnd_indices_set;
+};
+
+}  // namespace sampler
+}  // namespace pyg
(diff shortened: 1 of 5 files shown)


pyg_lib/csrc/sampler/cpu/indices_tracker.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <ATen/ATen.h>
+
+#include "parallel_hashmap/phmap.h"
+
+namespace pyg {
+namespace sampler {
+
+template <typename scalar_t>
+class IndicesTracker {
```
rusty1s, line 11: How is this different from `pyg::sampler::Mapper`? Can we unify the logic here?

pyg_lib/csrc/sampler/cpu/indices_tracker.h
```diff
@@ -0,0 +1,49 @@
...
+ public:
+  IndicesTracker(const scalar_t& size) : size(size) {
+    // TODO: add better switching threshold value mechanism?
+    use_vec = (size < 100000);
```
rusty1s, line 15: I feel this population threshold is a bit too large. I feel we can start with something like 100 or 1000 and see how it goes.

pyg_lib/csrc/sampler/cpu/indices_tracker.h
```diff
@@ -0,0 +1,49 @@
...
+    if (use_vec)
+      rnd_indices_vec.resize(size, 0);
+  }
+
+  bool tryInsert(const scalar_t& index) {
```
rusty1s, line 20:
```suggestion
  bool try_insert(const scalar_t& index) {
```

pyg_lib/csrc/sampler/cpu/indices_tracker.h
mszarma: I was doing some init analyses of using vector<char> vs unorder_set on neighborloader bench in pyg for ogbn-products and vector was always faster were max population for products was 10000+. Alternatively tested also vector<char> with using byte as storage for  bits and vector<bool> - vector<bool> was the slowest. Vector<char> as 8bit storage was also a little bit slower than vector<char>. In case of trade-off between memory usage vs perf, We can think about what to apply and what threshold to use.

pyg_lib/csrc/sampler/cpu/indices_tracker.h
mszarma: Main diff is set vs map usage (flat_hash set vs map). We can think how to unify it.

pyg_lib/csrc/sampler/cpu/indices_tracker.h
rusty1s: Yes, main point I wanted to make is if we shouldn't just use `std::vector` in the first place in the sampling without replacement logic. There is likely never a case where the population exceeds 10,000.

mszarma: "bench in pyg for ogbn-products and vector was always faster where max population for products was 10000+" -sorry, that sentence in misleading. I wanted to say that base on some initial research: vector was  faster than set for any population. And in case of products the populations max was going up to 10000+.

rusty1s (approved):
Thank you! <3

PR #99, opened 2022-09-09
fix(sampler): support empty `num_neighbors_dict`

mananshah99:


pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -309,14 +309,18 @@ sample(const std::vector<node_type>& node_types,
     const auto num_nodes = rowptr_dict.at(to_rel_type(k)).size(0) - 1;
     num_nodes_dict[!csc ? std::get<0>(k) : std::get<2>(k)] = num_nodes;
   }
+  for (const auto& kv : seed_dict) {
+    if (num_nodes_dict.find(kv.key()) == num_nodes_dict.end()) {
+      num_nodes_dict[kv.key()] = kv.value().size(0);
```
rusty1s, line 314: I assume this would need to be `seed.max() + 1`.

PR #106, opened 2022-09-19
Add hetero neighbor sampler benchmark

kgajdamo:
- Added hetero neighbor sampler benchmark to pyg-lib.
- Benchmark measures performance for the hetero_neighbor_sample from pyg-lib as well as hetero_neighbor_sample from pytorch_sparse.


benchmark/sampler/hetero_neighbor.py
```diff
@@ -0,0 +1,119 @@
...
+
+args = argparser.parse_args()
+
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+
+path = osp.join(osp.dirname(osp.realpath(__file__)), '../../data/OGB')
+transform = T.ToUndirected(merge=True)
+dataset = OGB_MAG(path, preprocess='metapath2vec', transform=transform)
+
+data = dataset[0].to(device, 'x', 'y')
```
rusty1s, line 46: Can be dropped

benchmark/sampler/hetero_neighbor.py
```diff
@@ -0,0 +1,119 @@
+import argparse
+import ast
+import os.path as osp
+import time
+
+import torch
+import torch_geometric.transforms as T
+import torch_sparse  # noqa
+from torch_geometric.datasets import OGB_MAG
+from torch_geometric.loader.utils import get_input_nodes
```
rusty1s, line 10: Can we avoid the dependency to PyG (besides for `OGB_MAG`)? I don't think it is desperately needed for benchmarking and these functions are not part of the public API and not stable.

benchmark/sampler/hetero_neighbor.py
```diff
@@ -0,0 +1,119 @@
...
+# TODO (kgajdamo): Enable CSR
+# argparser.add_argument('--csc', action='store_true')
+
+args = argparser.parse_args()
+
+device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+
+path = osp.join(osp.dirname(osp.realpath(__file__)), '../../data/OGB')
+transform = T.ToUndirected(merge=True)
+dataset = OGB_MAG(path, preprocess='metapath2vec', transform=transform)
```
rusty1s, line 44: We can move this to `pyg_lib.testing.withDataset`. WDYT? And then have it return a dictionary of `(rowptr, col)` entries?

benchmark/sampler/hetero_neighbor.py
```diff
@@ -0,0 +1,119 @@
...
+
+data = dataset[0].to(device, 'x', 'y')
+
+train_input_nodes = ('paper', data['paper'].train_mask)
+val_input_nodes = ('paper', data['paper'].val_mask)
+
+if args.do_train:
```
rusty1s, line 51: We can just iterate over the whole data once. I don't think we need to discriminate between train and validation nodes here: `input_nodes = {'paper': torch.arange(0, num_nodes)}`.

benchmark/sampler/hetero_neighbor.py
kgajdamo: Ok, I'll try

benchmark/sampler/hetero_neighbor.py
kgajdamo: Done.

benchmark/sampler/hetero_neighbor.py
rusty1s: Thanks! Can you clarify what is not working for sampling with replacement? Does it simply crash?

kgajdamo: I crashes in uniform_sample() -> add()

rusty1s: Thanks, will take a look!

rusty1s (approved):

rusty1s:
This is great, thank you very much!

It's interesting that `torch-sparse` performs better on `[-1]` neighborhoods. Wondering why this is the case.

kgajdamo:
> This is great, thank you very much!
> 
> It's interesting that `torch-sparse` performs better on `[-1]` neighborhoods. Wondering why this is the case.

Thanks @rusty1s for the updates. I have a question about your changes. In the hetero_neighbor.py file for the pytorch sampler case You put some declarations like node_types, edge_types etc inside the loop. These declarations make the measurement time greater and probably there is no need to declare them every time. Was it on purpose?
Another question is should I add dgl hetero neighbor sampler to the script?
And also I would like to ask why we take the matrix in csr format but the variable names are colptr and row?

kgajdamo:
> This is great, thank you very much!
> 
> It's interesting that `torch-sparse` performs better on `[-1]` neighborhoods. Wondering why this is the case.

Yes, pytorch_sparse sampler has better time when sample all one hop neighbors. Out of curiosity, I checked how it would look if num_neighbors=[-1, -1], and for this case pyg-lib is ~2 times faster. Here are results:
Screenshot 2022-09-23 at 16 38 23

rusty1s:
> Was it on purpose?

Yes, IMO that makes the comparison more fair as our Python wrapper around pyg-libs sample code does the same as well.

> Another question is should I add dgl hetero neighbor sampler to the script?

Yes, this would be very valuable.

> And also I would like to ask why we take the matrix in csr format but the variable names are colptr and row?

We need to use CSC here since that is the only format `torch-sparse` supports. Note that we take the CSR format of the transposed adjacency matrix with corresponds to CSC.

kgajdamo:
> > Was it on purpose?
> 
> Yes, IMO that makes the comparison more fair as our Python wrapper around pyg-libs sample code does the same as well.
> 
> > Another question is should I add dgl hetero neighbor sampler to the script?
> 
> Yes, this would be very valuable.
> 
> > And also I would like to ask why we take the matrix in csr format but the variable names are colptr and row?
> 
> We need to use CSC here since that is the only format `torch-sparse` supports. Note that we take the CSR format of the transposed adjacency matrix with corresponds to CSC.

Right, thank You for the detailed answer :)

PR #107, opened 2022-09-19
Build `*.so` extension on Mac instead of `*.dylib`

babymastodon:
By default, cmake created *.dylib, but python uses *.so. We either need to change cmake to write *.so, or change python to use *.dylib. I couldn't figure out how torch-sparse/scatter solve this problem (they seem to be creating *.so files), but happy to change the approach to align with the other projects, if you know.


rusty1s (approved):
Thanks!

PR #108, opened 2022-09-21
Require sorted neighborhoods according to time in temporal sampling

ZenoTan:
As discussed offline, we need to unify the implementation for temporal and non-temporal sampling.
This PR updates the neighbor sampling kernel by deciding the range of sampling based on seed node time. We assume we have sorted each neighborhood by time. This assumption will be guaranteed on pyg side in the follow-up PRs.


rusty1s (approved):

rusty1s:
This is amazing, thank you! Also added a note to the documentation.

PR #115, opened 2022-09-30
Reset `segment_matmul` feature granularity back to `1`

puririshi98:
In benchmarking found that the speedup of granularity 4 is inconsequential, even for shapes that don't require padding
considering the additional complexity  that padding adds and that it will inevitably take time to pad, it seems better to just go back to single granularity.

The reason the https://github.com/pyg-team/pyg-lib/pull/73 had such a speed boost was majority from the realization that the backward pass of other must still be for looped for optimal gpu utlization

download (1)


rusty1s (approved):
Thanks!

PR #121, opened 2022-10-08
fix(sampling): modify time satisfaction constraints to support equality

mananshah99:
This may be a misunderstanding on my part regarding temporal sampling logic, but it seems that destination nodes which have the same node time as the source node should satisfy the temporal constraint. This modification supports this, aligning with the corresponding logic in `torch_sparse` [here](https://github.com/rusty1s/pytorch_sparse/blob/master/csrc/cpu/neighbor_sample_cpu.cpp#L119-L130). 

If the `<` is intended, I'd be happy to understand more about why.


mananshah99:
Will update to support `<` and `<=`, as discussed offline. Feel free to ignore for now.

rusty1s:
Closing this for now. Feel free to re-open once we want to revisit this.

PR #124, opened 2022-10-13
Add `neighbor_kernel.h`, remove anonymous namespace

mananshah99:
It seems that the PyTorch register / dispatching logic is not fully reliable, which causes problems when depending on `neighbor.h` from external libraries. Until this is resolved in a more principled way, we introduce `neighbor_kernel.h` here for external libraries that wish to depend directly on `*_kernel` implementations, that avoid that logic.

### DIFF
--- a/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
+++ b/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
@@ -9,11 +9,10 @@
 #include "pyg_lib/csrc/sampler/subgraph.h"
 #include "pyg_lib/csrc/utils/cpu/convert.h"
 #include "pyg_lib/csrc/utils/types.h"
+#include "pyg_lib/csrc/sampler/cpu/neighbor_kernel.h"
 
 namespace pyg {
 namespace sampler {
-
-namespace {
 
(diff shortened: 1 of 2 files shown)


pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -554,8 +553,6 @@ hetero_neighbor_sample_kernel(
                   temporal_strategy);
 }
 
-}  // namespace
```
rusty1s, line 557: Only remove it for the two functions we expose?

pyg_lib/csrc/sampler/cpu/neighbor_kernel.h
```diff
@@ -0,0 +1,51 @@
+#include <ATen/ATen.h>
+#include <torch/library.h>
+
+#include "parallel_hashmap/phmap.h"
```
rusty1s, line 4: Only include the files needed?

rusty1s (approved):

pyg_lib/csrc/sampler/cpu/neighbor_kernel.h
mananshah99: Sorry, that was a leftover

PR #127, opened 2022-10-17
Make python tests part of CI

puririshi98:
although cpp tests are great, this test should be made a part of the CI as it catches bugs where it counts most, in the pythonic interface.
this error is in main:
```
Testing cpu:
test_segment_matmul_autograd passed!
Traceback (most recent call last):
  File "x.py", line 59, in <module>
    test_grouped_matmul_autograd(device_str)
  File "x.py", line 42, in test_grouped_matmul_autograd
    outs = pyg_lib.ops.grouped_matmul(inputs, others)
  File "/workspace/pyg-lib/pyg_lib/ops/__init__.py", line 33, in grouped_matmul
    outs = torch.ops.pyg.grouped_matmul(inputs, others)
  File "/opt/conda/lib/python3.8/site-packages/torch/_ops.py", line 175, in __call__
    return self._op(*args, **kwargs or {})
RuntimeError: mm(): functions with out=... arguments don't support automatic differentiation, but one of the arguments requires grad.
```

@dszwicht plz commit a fix for this issue


rusty1s (commented):
I don't think this is related to Python vs C++ tests. We currently do not test `grouped_matmul` backward in C++ tests, and autograd for `grouped_matmul` is currently disabled since I couldn't make `Autograd` dispatcher work for `TensorList`. Will look into it.

The main reason to block this PR right now is that I don't think we necessarily want to build the package twice. All Python ops should just call their respective C++ implementation anyway.

puririshi98:
oh that makes sense nvm

PR #128, opened 2022-10-19
Volta and fp32 support

puririshi98:
currently main only works on ampere using tf32. added fp32 support for ampere as well and support for volta devices

quick V100 HeteroLinear benchmark(vanilla algo runs out of mem for 1024 node types): 
image


rusty1s (approved):
Amazing, thank you!

PR #143, opened 2022-11-03
NeighborSample: allow different types for graph and timestamp data

ZenoTan:
Previously we use `scalar_t` for both graph and timestamp, this is not always good since most of the time the graph is small and we could hold it in int32 tensors. Note: we fix the temporal type to be int64, will address it in follow-up PRs if we find that we also need to dispatch temporal type other than graph type, which (nested dispatching) is not well supported by torch.


pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -217,7 +218,10 @@ sample(const at::Tensor& rowptr,
   AT_DISPATCH_INTEGRAL_TYPES(seed.scalar_type(), "sample_kernel", [&] {
     typedef std::pair<scalar_t, scalar_t> pair_scalar_t;
     typedef std::conditional_t<!disjoint, scalar_t, pair_scalar_t> node_t;
-    typedef NeighborSampler<node_t, scalar_t, replace, directed, return_edge_id>
+    // TODO(zeyuan): Do not force int64_t for time type.
+    typedef int64_t temporal_t;
```
mananshah99, line 222: Would it be better to check type equality (`static_assert` + `std::is_same`) and error out at compile time instead of implicitly promoting `temporal_t` to `int64_t`?

mananshah99 (commented):
LGTM, one comment about `temporal_t` implicit promotion

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
ZenoTan: The callers will only pass (dict of) tensors as args, so it might not be done in a static way. If we want to error out before pointer casting, I could add tensor type checking in the first place. wdyt?

rusty1s (approved):

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
mananshah99: I was instead thinking something like a static check of `temporal_t` within the template. Happy to chat further, not super important.

PR #146, opened 2022-11-16
`(segment|grouped)_matmul` implementation using MKL BLAS

dszwicht:
By default PyTorch is built with MKL BLAS support, we can take advantage of that and use [gemm_batch](https://www.intel.com/content/www/us/en/develop/documentation/onemkl-developer-reference-c/top/blas-and-sparse-blas-routines/blas-like-extensions/cblas-gemm-batch.html#cblas-gemm-batch) to implement `(segment|grouped)_matmul`.

`segment_matmul` C++ benchmarks (24 core machine) showed that the new version is on average 3.5 times faster than the previous one (with maximum speedup up to 6 times - note that chosen shapes are quite small, so speedup will be more impressive if a problem becomes more complex).

Additionally, HeteroLinear benchmarks were performed (thanks to @puririshi98 for sharing the benchmarks code):
segment_matmul_bench
Same benchmark with time in log2 scale:
segment_linear_bench_log2
We can see that `segment_matmul` execution path starts shining from `num_node_types=32` onward.


CONTRIBUTING.md
```diff
@@ -36,6 +36,14 @@ cmake .. -GNinja -DBUILD_TEST=ON -DBUILD_BENCHMARK=ON -DWITH_CUDA=ON -DCMAKE_CUD
 cmake --build .
 ```
 
+If you want to build a library with MKL BLAS support, you need to have MKL headers present on your machine.
+In a typical case, where MKL headers are installed via conda or pip, run following commands:
```
rusty1s, line 40:
```suggestion
In a typical case, where MKL headers are installed via `conda` or `pip`, run the following commands:
```

rusty1s (approved):
I think this looks great. Thanks for the effort!

Two questions:
* Can we get a better review in than mine from @pyg-team/intel-team?
* Can we enable it in our CPP tests to ensure that it is working correct?

dszwicht:
> Can we get a better review in than mine from @pyg-team/intel-team?

Sure, I'll ping someone.

> Can we enable it in our CPP tests to ensure that it is working correct?

Do you mean to enable it in the CI? 😉

rusty1s:
> Do you mean to enable it in the CI? 😉

Yes.

yanbing-j:
LGTM. Thanks for your hard work!
I have question about the improvement in E2E, do you verify this in a real benchmark?

dszwicht:
> > Do you mean to enable it in the CI? 😉
> 
> Yes.

Just added. Please take another look. 😉

dszwicht:
> LGTM. Thanks for your hard work! I have question about the improvement in E2E, do you verify this in a real benchmark?

Training and inference benchmarks, unfortunately, do not cover the `segment_matmul` execution path in HeteroLinear or RGCNConv.

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,27 +1,274 @@
...
+            });
+      });
 }
 
 std::vector<at::Tensor> grouped_matmul_kernel(const at::TensorList input,
                                               const at::TensorList other) {
-  std::vector<at::Tensor> out(input.size());
-  for (size_t i = 0; i < input.size(); ++i)
-    out[i] = input[i].new_empty({input[i].size(0), other[i].size(-1)});
+  std::vector<at::Tensor> input_contig;
```
mingfeima, line 260: the constructor of `at::Tensor` is actually very heavy, consider use a meta struct (you only need dataptr, dims, sizes, strides here).

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,27 +1,274 @@
...
+  std::vector<at::Tensor> other_contig;
+  std::vector<at::Tensor> out;
+
+  for (size_t i = 0; i < input.size(); ++i) {
+    input_contig.push_back(input[i].contiguous());
+    other_contig.push_back(other[i].contiguous());
```
mingfeima, line 266: use `embrace_back` and `reserve` for std::vectors is usually a good practice, otherwise if the list if long, you will get a lot of memory allocation in your vtune profile.

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -31,16 +278,18 @@ at::Tensor segment_matmul_kernel(const at::Tensor& input,
...
 
   auto outs = out.split_with_sizes(/*split_size=*/sizes, /*dim=*/0);
   for (auto& out_part : outs) {
     out_part.resize_(0);
   }
 
   grouped_matmul_out_kernel(
-      input.contiguous().split_with_sizes(/*split_size=*/sizes, /*dim=*/0),
-      other.contiguous().split(/*split_size=*/1, /*dim=*/0), outs);
+      input_contig.split_with_sizes(/*split_size=*/sizes, /*dim=*/0),
```
mingfeima, line 291: `split_with_sizes` will create new tensor views, again it is `at::Tensor` constructor issue.
Notice that this is a very heavy struct, we should avoid construct/deconstruct heavy struct as much as possible.

mingfeima (changes requested):
Do we have test cases for it?

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,27 +1,274 @@
...
+                                    const std::vector<int>& ks,
+                                    const std::vector<int>& ld_src0,
+                                    const std::vector<int>& ld_src1,
+                                    const std::vector<int>& ld_dst,
+                                    const std::vector<int>& group_sizes,
+                                    const std::vector<scalar_t>& alpha,
+                                    const std::vector<scalar_t>& beta,
+                                    const int group_count) {
+  if (group_count > 1) {
+#pragma omp parallel for schedule(dynamic)
```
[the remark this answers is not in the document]
dszwicht: Done.

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
dszwicht: Done.

dszwicht:
> Do we have test cases for it?

Yes, [here](https://github.com/pyg-team/pyg-lib/blob/master/test/csrc/ops/test_matmul.cpp) and [here](https://github.com/pyg-team/pyg-lib/blob/master/test/ops/test_matmul.py).

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,27 +1,274 @@
...
+                           const float** a_array,
+                           const int* lda_array,
+                           const float** b_array,
+                           const int* ldb_array,
+                           const float* beta_array,
+                           float** c_array,
+                           const int* ldc_array,
+                           const int group_count,
+                           const int* group_size) {
+  std::vector<CBLAS_TRANSPOSE> transa(group_count, CblasNoTrans);
```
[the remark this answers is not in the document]
dszwicht: That was of course designed with contiguous data in mind (with some input requiring transposition here, some negative impact on gemm_batch computation time will be observed - unless every data is transposed, then I guess CBlasColMajor layout should be chosen).

Anyway, as stated in other comments, I think I'll stick with a contiguous layout, as it is the common case.

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
dszwicht: Good point 😉 I think I can get rid of `split_with_sizes` in the MKL execution path by changing leading dimensions.

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,27 +1,274 @@
...
+
+void grouped_matmul_out_kernel_at_impl(const std::vector<at::Tensor> input,
+                                       const std::vector<at::Tensor> other,
+                                       std::vector<at::Tensor> out) {
+  for (size_t i = 0; i < out.size(); ++i) {
     at::matmul_out(const_cast<at::Tensor&>(out[i]), input[i], other[i]);
+  }
+}
+
+void grouped_matmul_out_kernel(const std::vector<at::Tensor> input,
```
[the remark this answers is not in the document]
dszwicht: Good point 👍 This is something that definitely could be covered here, but with [current usage](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/dense/linear.py#L269) of `segment_matmul` (transposed data not expected), I think I will stick with the contiguous layout (and improve this later if the need will arise).

dszwicht:
@mingfeima Please take another look. 😉

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,46 +1,412 @@
...
+                                    const int group_count,
+                                    const std::vector<int>& group_sizes) {
+  int64_t work_size = 0;
+  for (size_t i = 0; i < group_count; ++i) {
+    work_size += ks[i] * group_sizes[i];
+  }
+  const int64_t grain_size = at::internal::GRAIN_SIZE / work_size;
+
+  if (group_count > 1) {
+    at::parallel_for(0, group_count, grain_size, [&](size_t beg, size_t end) {
```
mingfeima, line 137: so this one is a nested omp loop: calling mkl inside an openmp loop, the default behavior of mkl is to go sequential inside here since, so only the outer openmp loop will go parallel. Is this your expectation ?

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,46 +1,412 @@
...
+        const scalar_t** src1_ptrs_local = src1_ptrs + offset;
+        scalar_t** dst_ptrs_local = dst_ptrs + offset;
+        mkl_blas_gemm_batched(&ms[i], &ns[i], &ks[i], &alpha[i],
+                              src0_ptrs_local, &ld_src0[i], src1_ptrs_local,
+                              &ld_src1[i], &beta[i], dst_ptrs_local, &ld_dst[i],
+                              1, &group_sizes[i]);
+      }
+    });
+  } else {
+    at::parallel_for(
```
mingfeima, line 152: why do we still need `at::parallel_for` here when `group_count == 1`? let mkl itself decide when to go parallel won't do the job ?

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,46 +1,412 @@
...
+        c10::guts::if_constexpr<WITH_MKL_BLAS() && AT_MKL_ENABLED() &&
+                                is_blas_library_type<scalar_t>::value>(
+            [&](auto _) {
+              segment_matmul_out_kernel_mkl_impl(input_contig, other_contig,
+                                                 out, sizes);
+            },
+            [&](auto _) {
+              auto outs = out.split_with_sizes(/*split_size=*/sizes, /*dim=*/0);
+              for (auto& out_part : outs) {
+                out_part.resize_(0);
```
mingfeima, line 402: `resize_(0)` is kind of strange, why do we need it?

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -1,46 +1,412 @@
...
+  std::vector<at::Tensor> input_contig;
+  std::vector<at::Tensor> other_contig;
+  std::vector<at::Tensor> out;
+
+  input_contig.reserve(n_matrices);
+  other_contig.reserve(n_matrices);
+  out.reserve(n_matrices);
+
+  for (size_t i = 0; i < n_matrices; ++i) {
+    input_contig.emplace_back(input[i].contiguous());
```
mingfeima, line 265: this is OK if we expect the `input` and `other` are given in the contiguous format already.
if not, say `other` is transposed, we need to change the logic here.

mingfeima (approved):

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
dszwicht: After splitting the `other` tensor, we get shapes like (1, K, M), so the output is expected to be in shape (1, N, M). That is not happening and a warning is generated. The code here is to fix that behavior (recommended in the warning). On the 2nd look now, it looks weird to me as well, so I changed it to a more intuitive `unsqueeze_`.

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
dszwicht: No nested loops here. I call this function, only if PyTorch is linking to MKL sequential version. In that case, `at::parallel_for` is the only loop.

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
dszwicht: As in the comment above, this code will be executed only if we use MKL sequential version, in that case, we want to provide simple parallelization.

PR #161, opened 2022-12-08
Added `bias` term to `grouped_matmul` and `segment_matmul`

puririshi98:
for https://github.com/pyg-team/pytorch_geometric/pull/6178

### DIFF
--- a/pyg_lib/ops/__init__.py
+++ b/pyg_lib/ops/__init__.py
@@ -6,7 +6,7 @@ from torch import Tensor
 from .scatter_reduce import fused_scatter_reduce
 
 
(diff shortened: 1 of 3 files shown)


pyg_lib/ops/__init__.py
```diff
@@ -6,7 +6,8 @@
 from .scatter_reduce import fused_scatter_reduce
 
 
-def grouped_matmul(inputs: List[Tensor], others: List[Tensor]) -> List[Tensor]:
+def grouped_matmul(inputs: List[Tensor], others: List[Tensor],
```
rusty1s, line 9: I think I would like to keep `grouped_matmul` and `segment_matmul` interfaces consistent. We can either add this functionality to `segment_matmul` as well or solely perform adding the bias on PyG side (my current preference).

rusty1s (approved):

PR #167, opened 2022-12-30
Added `--write-csv` and `--libraries` parameters to `neighbor.py` and `hetero_neighbor.py` benchmarks

OlhaBabicheva:
Added --write-csv (saves benchmark results as csv file) and --libraries (determines which libraries will be used in benchmark) parameters to neighbor.py and hetero_neighbor.py.
Output for command python hetero_neighbor.py --write-csv :
image


rusty1s (approved):
Thank you!

PR #172, opened 2023-01-10
`(segment|grouped)_matmul` improvements

dszwicht:
Added several fixes and improvements for `(segment|grouped)_matmul`:
* PyTorch stores shape information in `int64_t` variables, whereas MKL uses regular `int`. This means that tensor size can exceed the maximum value acceptable by MKL - in that case we need to use the ATen execution path.
* `config.h` file was removed and added to the .gitignore list - it is automatically generated during the build from the `config.h.in`.
* grain size in MKL sequential execution path was improved, resulting in an average 53% speedup.

### DIFF
--- a/pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
+++ b/pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
@@ -1,3 +1,5 @@
+#include <algorithm>
+#include <climits>
 #include <numeric>
 #include <tuple>
 #include <type_traits>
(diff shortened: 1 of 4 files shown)


pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -113,14 +115,46 @@ using is_blas_library_type =
                            std::is_same<scalar_t, float>::value ||
                                std::is_same<scalar_t, double>::value>;
 
+template <typename scalar_t>
+bool mkl_path_available() {
+  if (WITH_MKL_BLAS() && AT_MKL_ENABLED() &&
+      is_blas_library_type<scalar_t>::value) {
+    return true;
+  }
+  return false;
```
rusty1s, line 124:
```suggestion
  return (WITH_MKL_BLAS() && AT_MKL_ENABLED() &&
      is_blas_library_type<scalar_t>::value);
```

pyg_lib/csrc/ops/cpu/matmul_kernel.cpp
```diff
@@ -113,14 +115,46 @@ using is_blas_library_type =
...
+}
+
+bool mkl_path_possible(const at::IntArrayRef& sizes, int64_t n, int64_t k) {
+  const int64_t limit = INT_MAX;
```
rusty1s, line 128: I feel like a simple for-loop would be easier to read, but no strong pref.

rusty1s (approved):
Thanks! This is cool.

PR #183, opened 2023-01-18
`group_matmul` support for matmuls of lists of 3D tensors. needed for `HGTConv`

puririshi98:
https://github.com/pytorch/pytorch/issues/91904
needed for: https://github.com/pyg-team/pytorch_geometric/pull/6178

### DIFF
--- a/pyg_lib/ops/__init__.py
+++ b/pyg_lib/ops/__init__.py
@@ -41,10 +41,10 @@ def grouped_matmul(inputs: List[Tensor], others: List[Tensor],
         inputs = torch.nested.as_nested_tensor(inputs).contiguous()
         others = torch.nested.as_nested_tensor(others).contiguous()
         if inputs.dim() == 4 or others.dim() == 4:
-          # bmm only works on lists of 2D tensors
-          outs = torch.matmul(inputs, others).contiguous()
+            # bmm only works on lists of 2D tensors
+            outs = torch.matmul(inputs, others).contiguous()
         else:
-          outs = torch.bmm(inputs, others).contiguous()
+            outs = torch.bmm(inputs, others).contiguous()
         outs = list(outs.unbind())
     else:
         input_req_grad = any([i.requires_grad for i in inputs])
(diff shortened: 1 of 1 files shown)


pyg_lib/ops/__init__.py
```diff
@@ -40,7 +40,11 @@ def grouped_matmul(inputs: List[Tensor], others: List[Tensor],
     if int(major_vers) >= 2 or int(minor_vers) >= 14:
         inputs = torch.nested.as_nested_tensor(inputs).contiguous()
         others = torch.nested.as_nested_tensor(others).contiguous()
-        outs = torch.bmm(inputs, others).contiguous()
+        if inputs.dim() == 4 or others.dim() == 4:
+            # bmm only works on lists of 2D tensors
+            outs = torch.matmul(inputs, others).contiguous()
```
rusty1s, line 45: Do `matmul` and `bmm` use different code-paths? Is it slower to use `matmul`? Can you shed some light here?

rusty1s (approved):

pyg_lib/ops/__init__.py
puririshi98: i believe the matmul option is using a slower code path under the hood, but since i needed to deal w/ 3D tensor lists for HGTConv it was the only option. if this is undesirable, then i would have to do HGTConv w/ segment matmul only, which requires padding under the hood. thoughts on which you would like to see? if you'd like i can try to benchmark the torch >= 1.14 grouped matmul path agaisnt the torch < 1.14 segment matmul path to see how their speed compares. that can inform our decision

puririshi98: unless you see a way to shape tensors [here](https://github.com/pyg-team/pytorch_geometric/blob/hgtconv_pyglib/torch_geometric/nn/conv/hgt_conv.py#L298-L327) to be a list of 2D tensors instead of 3D. I tried a few ways but ended up with the wrong shapes or some other issue

rusty1s: Never really thought about this use-case. I am happy to merge this now, and potentially revisit the internal code path later.

PR #184, opened 2023-01-19
`ops/matmul`: Only check cuda device properties once

puririshi98:


rusty1s (approved):

pyg_lib/csrc/ops/cuda/matmul_kernel.cu
```diff
@@ -103,30 +103,28 @@ int shared_memory_for_kernel() {
...
-    return -1;
+    throw std::runtime_error(cudaGetErrorString(result));
   }
-
-  return properties.sharedMemPerMultiprocessor;
+  return properties;
+}
+cudaDeviceProp props;
+if (torch::cuda::is_available()) {
+  props = get_dev_prop();
```
rusty1s, line 122: Doesn't this get invoked directly on import? Wouldn't it be more reasonable to cache it instead, e.g., via `static`?

puririshi98: yeah i noticed thats why the CI was failing. will fix

puririshi98:
got it wokring:
test/ops/test_matmul.py::test_segment_matmul_autograd[device0] PASSED
test/ops/test_matmul.py::test_segment_matmul_autograd[device1] PASSED
test/ops/test_matmul.py::test_grouped_matmul_autograd[device0] PASSED
test/ops/test_matmul.py::test_grouped_matmul_autograd[device1] PASSED

PR #192, opened 2023-01-31
Fix `index_sort`

dszwicht:
It appears that `vectorized_copy` was causing memory errors in some workloads. This PR fixes these errors.


dszwicht:
@rusty1s Could you please take a look at this fix? I would like it to be merged before [pytorch_geometric:6554](https://github.com/pyg-team/pytorch_geometric/pull/6554)

rusty1s (approved):

PR #197, opened 2023-02-10
Adding option to return info about sampled graph

mszarma:
It's enabling the hierarchical tensor usage
and significant performance improvement

PyG part: https://github.com/pyg-team/pytorch_geometric/pull/6661


pyg_lib/sampler/__init__.py
```diff
@@ -62,17 +63,21 @@ def neighbor_sample(
         return_edge_id (bool, optional): If set to :obj:`False`, will not
             return the indices of edges of the original graph.
             (default: :obj: `True`)
-
+        return_sampled_info (bool): If set to :obj:`True`, will return information about
+            amount of sampled nodes and edges per layer.
```
rusty1s, line 67:
```suggestion
        return_sampled_info (bool): If set to :obj:`True`, will return information about
            the amount of sampled nodes and edges per layer.
```

pyg_lib/sampler/__init__.py
```diff
@@ -62,17 +63,21 @@ def neighbor_sample(
...
+            (default: :obj: `False`)
     Returns:
-        (torch.Tensor, torch.Tensor, torch.Tensor, Optional[torch.Tensor]):
+        (torch.Tensor, torch.Tensor, torch.Tensor, Optional[torch.Tensor],
+        Optional[torch.Tensor], Optional[torch.Tensor]):
         Row indices, col indices of the returned subtree/subgraph, as well as
         original node indices for all nodes sampled.
-        In addition, may return the indices of edges of the original graph.
+        In addition, may return the indices of edges of the original graph,
+        as well as information about sampled amount of nodes and edges per layer.
```
rusty1s, line 75:
```suggestion
        as well as information about the sampled amount of nodes and edges per layer.
```

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -24,13 +24,17 @@ template <typename node_t,
...
 class NeighborSampler {
  public:
   NeighborSampler(const scalar_t* rowptr,
                   const scalar_t* col,
                   const std::string temporal_strategy)
-      : rowptr_(rowptr), col_(col), temporal_strategy_(temporal_strategy) {
+      : rowptr_(rowptr),
+        col_(col),
+        temporal_strategy_(temporal_strategy),
+        num_neighbors_counter(0) {
```
rusty1s, line 37: can we rename this to `curr_hop` or something like this?

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -93,6 +97,10 @@ class NeighborSampler {
     }
   }
 
+  std::vector<scalar_t> get_sampled_edges_num() {
```
rusty1s, line 100:
```suggestion
  std::vector<scalar_t> get_sampled_info() {
```

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -170,6 +178,9 @@ class NeighborSampler {
       out_global_dst_nodes.push_back(global_dst_node);
     }
     if (save_edges) {
+      if (return_sampled_info) {
+        sampled_edges_per_num_neighbors[num_neighbors_counter]++;
```
rusty1s, line 182:
```suggestion
        sampled_edges_per_hop[num_neighbors_counter]++;
```

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -277,27 +309,41 @@ sample(const at::Tensor& rowptr,
         }
       }
       begin = end, end = sampled_nodes.size();
+      if (return_sampled_info)
+        nodes_num_per_neighbor.push_back(end - begin);
```
rusty1s, line 313:
```suggestion
        num_nodes_per_hop.push_back(end - begin);
```

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -437,11 +495,16 @@ sample(const std::vector<node_type>& node_types,
         std::tie(begin, end) = slice_dict.at(src);
 
         if (!time_dict.has_value() || !time_dict.value().contains(dst)) {
+          if (!disjoint && return_sampled_info)
```
rusty1s, line 498: Why do we need to gate this behind `disjoint`?

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -475,53 +551,129 @@ sample(const std::vector<node_type>& node_types,
...
-    return sample<false, true, false, false>(__VA_ARGS__);                \
-  if (!replace && !directed && disjoint && return_edge_id)                \
-    return sample<false, false, true, true>(__VA_ARGS__);                 \
-  if (!replace && !directed && disjoint && !return_edge_id)               \
-    return sample<false, false, true, false>(__VA_ARGS__);                \
-  if (!replace && !directed && !disjoint && return_edge_id)               \
-    return sample<false, false, false, true>(__VA_ARGS__);                \
-  if (!replace && !directed && !disjoint && !return_edge_id)              \
-    return sample<false, false, false, false>(__VA_ARGS__);
+#define DISPATCH_SAMPLE(replace, directed, disjount, return_edge_id, \
```
rusty1s, line 570: Not a real expert on this one, but is there some way to auto-generate this? This gets pretty long :D

rusty1s (commented):
I think code-wise this looks all good. Remaining questions would be
* Can we clean-up dispatching logic? Do you have any solution for this?
* Would prefer some renaming ala `num_nodes_per_hop` or similar
* I'm not totally sure why we gate behind `disjoint` in the heterogeneous case but not in the homogeneous one.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
mszarma: Please look to the main reply.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
mszarma: Please look to the main reply.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
mszarma: Done

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
mszarma: Done

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
mszarma: Done

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
mszarma: Done

pyg_lib/sampler/__init__.py
mszarma: Done

pyg_lib/sampler/__init__.py
mszarma: Done

mszarma:
> I think code-wise this looks all good. Remaining questions would be
> 
> * Can we clean-up dispatching logic? Do you have any solution for this?
> * Would prefer some renaming ala `num_nodes_per_hop` or similar
> * I'm not totally sure why we gate behind `disjoint` in the heterogeneous case but not in the homogeneous one.

Thanks you @rusty1s  for the review.

> * Can we clean-up dispatching logic? Do you have any solution for this?
I would recommend to put it to the PyG backlog and then think how to tackle that. Would prefer not to mix it. Move forward the feature part separately and then as separate task think how to refactor that dispatching logic. My gut feeling is that it may need a major refactor.
> * Would prefer some renaming ala `num_nodes_per_hop` or similar
That's a good idea. Done. 
> * I'm not totally sure why we gate behind `disjoint` in the heterogeneous case but not in the homogeneous one.
The difference in hetero flow is that we are using the dict/Phmap. The issue is in usage of node_t: https://github.com/pyg-team/pyg-lib/blob/master/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp#L220. In the disjoint hetero flow compiler point out that phmap cannot match/use the pair type. I didn't investigated usage of disjoint/temporal flow so my recommendation would be to enable feature firstly without disjoint support. WDYT @rusty1s ?

rusty1s (approved):

PR #204, opened 2023-02-24
chore: create `pyproject.toml`

SauravMaheshkar:
Similar to (https://github.com/pyg-team/pytorch_geometric/pull/6431)

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -19,6 +19,10 @@ runs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ inputs.python-version }}
+        check-latest: true
+        cache: pip
+        cache-dependency-path: |
+          setup.py
 
     - name: Free up disk space
       if: ${{ runner.os == 'Linux' }}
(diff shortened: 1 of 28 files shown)


SauravMaheshkar:
It seems like the CI is defaulting to `pyproject.toml` for editable installs.

[Source](https://github.com/pyg-team/pyg-lib/actions/runs/4262515982/jobs/7418111896#step:4:204)

rusty1s:
Mh, this is strange. I assume this fails now as we don't specify torch as a required dependency in `setup.py`?

SauravMaheshkar:
> Mh, this is strange. I assume this fails now as we don't specify torch as a required dependency in `setup.py`?

Tried that in [00e7509](https://github.com/pyg-team/pyg-lib/pull/204/commits/00e75095e2653c88c71216a37c3145a94b513fa7), didn't work 😞

SauravMaheshkar:
@rusty1s I figured it out !!! Turns out we needed to use the `--no-use-pep517` flag during build. 

For more context, refer to [PEP 517](https://peps.python.org/pep-0517/)

rusty1s:
Mh, interesting. Does that mean that all users would now also install the package via `pip install --no-use-pep517`? Personally, I would like to avoid this :D

SauravMaheshkar:
@rusty1s I think if we ensure that we're running on the latest version of pip might help. This would only involve a change in `actions/setup/action.yml` (a simple addition of `pip install -U pip setuptools wheel`).

SauravMaheshkar:
> @rusty1s I think if we ensure that we're running on the latest version of pip might help. This would only involve a change in `actions/setup/action.yml` (a simple addition of `pip install -U pip setuptools wheel`).

@rusty1s that hypothesis failed e194a8fca0b6cc4cf53068ac8920ec7e342c104c

rusty1s:
Mh, what a bummer. Let me do some research on this as well.

SauravMaheshkar:
Closing this for now, (atleast until the `--no-use-pep517` flag is made default by the latest versions)

PR #208, opened 2023-02-28
feat(pre-commit): add `clang-format` config to pre-commit-config

SauravMaheshkar:
This PR aims to cleanup the repository structure by moving the `.clang-format` file into the `pre-commit-config.yaml`.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,1 +1,0 @@
-BasedOnStyle: Chromium
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,3 +48,4 @@ repos:
     hooks:
       - id: clang-format
         name: Format C++ code
+        args: [--style, "{BasedOnStyle: Chromium}"]


SauravMaheshkar:
I'm assuming we'd want to skip changelog for this PR

rusty1s:
Thanks for the PR. If we only set this to the `pre-commit` config, I would assume that it is no longer used by local linters? Wouldn't this be a problem?

SauravMaheshkar:
Yes moving this to the `pre-commit` config would make it no longer accessible by local linters, and the default style for `clang-format` is `llvm`

rusty1s:
Mh, so shall we keep the old `.clang-format` as well?

SauravMaheshkar:
If we want to keep the old `.clang-format` as well then we probably should change the `pre-commit` config from `args: [--style, "{BasedOnStyle: Chromium}"]` to `--style=file`.

rusty1s:
That would work for me.

rusty1s (approved):

PR #215, opened 2023-03-16
Hetero neighbor sampler multithreading

kgajdamo:
The main roadblocks in the context of sampler parallelization is the mapper and the sampler. However, in the case of hetero sampling, there are separate mappers for each dst node type and separate samplers for each edge type. Therefore, to avoid race condition we can use parallelization per dst node type, i.e. each thread is assigned an edge types with unique dst node type that it is working on.

Example sampling times obtained using  [hetero_neighbor.py](https://github.com/pyg-team/pyg-lib/blob/master/benchmark/sampler/hetero_neighbor.py) benchmark:

hetero-sampler-icx

Example end-to-end times obtained using [inference_benchmark.py](https://github.com/pyg-team/pytorch_geometric/blob/master/benchmark/inference/inference_benchmark.py):

e2e-icx

### DIFF
--- a/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
+++ b/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
@@ -435,8 +435,7 @@ sample(const std::vector<node_type>& node_types,
           threads_edge_types.push_back({k});
       }
     }
-    if (!parallel) {
-      // If not parallel then one thread handles all edge types
+    if (!parallel) {  // If not parallel then one thread handles all edge types.
       threads_edge_types.push_back({edge_types});
     }
 
@@ -483,10 +482,8 @@ sample(const std::vector<node_type>& node_types,
       phmap::flat_hash_map<node_type, std::vector<node_t>>
           dst_sampled_nodes_dict;
       if (parallel) {
-        for (const auto& k : threads_edge_types) {
-          dst_sampled_nodes_dict[!csc ? std::get<2>(k[0])
-                                      : std::get<0>(
-                                            k[0])];  // initialize empty vector
+        for (const auto& k : threads_edge_types) {  // Intialize empty vectors.
+          dst_sampled_nodes_dict[!csc ? std::get<2>(k[0]) : std::get<0>(k[0])];
         }
       }
       at::parallel_for(
--- a/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
+++ b/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
@@ -538,15 +538,12 @@ sample(const std::vector<node_type>& node_types,
                     std::back_inserter(sampled_nodes_dict[dst_sampled.first]));
         }
       }
-      at::parallel_for(0, node_types.size(), 1, [&](size_t _s, size_t _e) {
-        for (auto j = _s; j < _e; j++) {
-          const auto& k = node_types[j];
-          slice_dict[k] = {slice_dict.at(k).second,
-                           sampled_nodes_dict.at(k).size()};
-          num_sampled_nodes_per_hop_map.at(k).push_back(
-              slice_dict.at(k).second - slice_dict.at(k).first);
-        }
-      });
+      for (const auto& k : node_types) {
+        slice_dict[k] = {slice_dict.at(k).second,
+                         sampled_nodes_dict.at(k).size()};
+        num_sampled_nodes_per_hop_map.at(k).push_back(slice_dict.at(k).second -
+                                                      slice_dict.at(k).first);
+      }
     }
 
     for (const auto& k : node_types) {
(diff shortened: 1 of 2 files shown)


mingfeima:
The speedup does not see to be very much, do we have analysis on what is the bottleneck ?

kgajdamo:
> The speedup does not see to be very much, do we have analysis on what is the bottleneck ?

I think the reason is that the number of threads is limited to the number of dst nodes. As many different dst nodes there are as many threads. For the ogbn-mag dataset, this will be 4.

mingfeima:
> > The speedup does not see to be very much, do we have analysis on what is the bottleneck ?
> 
> I think the reason is that the number of threads is limited to the number of dst nodes. As many different dst nodes there are as many threads. For the ogbn-mag dataset, this will be 4.

The code generally LGTM, but i am still curious on the limited speedup. You may try from two aspects:
* can we shift to another dataset with more dst nodes.
* check VTune log on ogbn-mag dataset.

kgajdamo:
> > > The speedup does not see to be very much, do we have analysis on what is the bottleneck ?
> > 
> > 
> > I think the reason is that the number of threads is limited to the number of dst nodes. As many different dst nodes there are as many threads. For the ogbn-mag dataset, this will be 4.
> 
> The code generally LGTM, but i am still curious on the limited speedup. You may try from two aspects:
> 
> * can we shift to another dataset with more dst nodes.
> * check VTune log on ogbn-mag dataset.

Sure, I'll check that in VTune.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -453,42 +480,73 @@ sample(const std::vector<node_type>& node_types,
...
+            }
+          });
+
+      if (parallel) {
+        for (auto& dst_sampled : dst_sampled_nodes_dict) {
+          std::copy(dst_sampled.second.begin(), dst_sampled.second.end(),
+                    std::back_inserter(sampled_nodes_dict[dst_sampled.first]));
+        }
       }
+      at::parallel_for(0, node_types.size(), 1, [&](size_t _s, size_t _e) {
```
rusty1s, line 541: Does this bring any gain? We are not doing heavy work here, so the overhead of threading might not be worth it?

rusty1s (commented):
Looks good. A few minor comments.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -435,12 +460,14 @@ sample(const std::vector<node_type>& node_types,
         if (seed_time_dict.has_value()) {
           const at::Tensor& seed_time = seed_time_dict.value().at(kv.key());
           const auto seed_time_data = seed_time.data_ptr<scalar_t>();
+          seed_times.reserve(seed_times.size() + seed.numel());
```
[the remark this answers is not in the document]
kgajdamo: The reserve function does not reserve memory from the end, but from the beginning. So the size should include the elements that have already been added to the vector:
Screenshot 2023-05-02 at 12 07 24

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
kgajdamo: But maybe to be sure, I'll take 10 measurements and average the results

kgajdamo: The average of 10 measurements is as follows:
Screenshot 2023-05-03 at 15 38 45

Looks like the results are almost the same. Should I remove it then? WDYT?

rusty1s: I would be in favor of removing it, as it doesn't look that useful.

kgajdamo: Sure

mingfeima (approved):

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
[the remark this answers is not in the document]
rusty1s: Yeah, I know. It's just that `seed_times` is empty there anyway, but I guess your approach is indeed more correct.

rusty1s (approved):

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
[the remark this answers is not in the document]
kgajdamo: But all the code is in a for loop `(const auto& kv : seed_dict)`
so it's possible that new elements will be added in subsequent passes.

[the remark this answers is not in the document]
rusty1s: Got it.

PR #222, opened 2023-04-04
MKL RandIntEngine

OlhaBabicheva:
- Added fill_with_ints function (it uses mkl library in order to generate ints) to RandintEngine class
Results obtained using neighbor.py benchmark (average time from 15 runs):

Results for changed code:

With mkl:
image

Without mkl:
image

Results for original code:
image

### DIFF
--- a/pyg_lib/csrc/random/cpu/rand_engine.h
+++ b/pyg_lib/csrc/random/cpu/rand_engine.h
@@ -3,6 +3,10 @@
 #include <ATen/ATen.h>
 
 #include <limits.h>
+#include "pyg_lib/csrc/config.h"
+#if WITH_MKL_BLAS()
+#include "mkl_vsl.h"
+#endif
 
 namespace pyg {
 namespace random {
@@ -99,18 +103,41 @@ class PrefetchedRandint {
 template <typename T>
 class RandintEngine {
  public:
-  RandintEngine() : prefetched_(RAND_PREFETCH_SIZE, RAND_PREFETCH_BITS) {}
+  RandintEngine() {
+    #if WITH_MKL_BLAS()
+    vslNewStream(&stream, VSL_BRNG_MT19937, 1);
+    #endif
+  }
+  ~RandintEngine() {
+    #if WITH_MKL_BLAS()
+    vslDeleteStream(&stream);
+    #endif
+  }
 
   // Uniform random number within range [beg, end)
   T operator()(T beg, T end) {
     TORCH_CHECK(beg < end, "Randint engine illegal range");
 
     T range = end - beg;
+    if(!start_prefetch) {
+      new (&prefetched_) PrefetchedRandint(RAND_PREFETCH_SIZE, RAND_PREFETCH_BITS);
+      start_prefetch = true;
+    }
     return prefetched_.next(range) + beg;
+  }
+
+  void generate_ints(T beg, T end, T count, int* arr) {
+    #if WITH_MKL_BLAS()
+    viRngUniform(VSL_RNG_METHOD_UNIFORM_STD, stream, count, arr, beg, end);
+    #endif
   }
 
  private:
   PrefetchedRandint prefetched_;
+  bool start_prefetch = false;
+  #if WITH_MKL_BLAS()
+  VSLStreamStatePtr stream;
+  #endif
 };
 
 /**
--- a/pyg_lib/csrc/random/cpu/rand_engine.h
+++ b/pyg_lib/csrc/random/cpu/rand_engine.h
@@ -119,22 +119,27 @@ class RandintEngine {
     TORCH_CHECK(beg < end, "Randint engine illegal range");
 
     T range = end - beg;
-    if(!start_prefetch) {
-      new (&prefetched_) PrefetchedRandint(RAND_PREFETCH_SIZE, RAND_PREFETCH_BITS);
-      start_prefetch = true;
+    if(!prefetch_initialized) {
+      prefetched_ = PrefetchedRandint(RAND_PREFETCH_SIZE, RAND_PREFETCH_BITS);
+      prefetch_initialized = true;
     }
     return prefetched_.next(range) + beg;
   }
 
-  void generate_ints(T beg, T end, T count, int* arr) {
+  void fill_with_ints(T beg, T end, T count, int* ptr) {
     #if WITH_MKL_BLAS()
-    viRngUniform(VSL_RNG_METHOD_UNIFORM_STD, stream, count, arr, beg, end);
+    viRngUniform(VSL_RNG_METHOD_UNIFORM_STD, stream, count, ptr, beg, end);
+    #else
+    for(size_t i = 0; i < count; ++i){
+      *ptr = (*this)(beg, end);
+      ++ptr;
+    }
     #endif
   }
 
  private:
   PrefetchedRandint prefetched_;
-  bool start_prefetch = false;
+  bool prefetch_initialized = false;
   #if WITH_MKL_BLAS()
   VSLStreamStatePtr stream;
   #endif
--- a/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
+++ b/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
@@ -10,6 +10,7 @@
 #include "pyg_lib/csrc/sampler/subgraph.h"
 #include "pyg_lib/csrc/utils/cpu/convert.h"
 #include "pyg_lib/csrc/utils/types.h"
+#include "pyg_lib/csrc/config.h"
 
 namespace pyg {
 namespace sampler {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `bias` term to `grouped_matmul` and `segment_matmul` ([#161](https://github.com/pyg-team/pyg-lib/pull/161))
 - Added `sampled_op` implementation ([#156](https://github.com/pyg-team/pyg-lib/pull/156), [#159](https://github.com/pyg-team/pyg-lib/pull/159), [#160](https://github.com/pyg-team/pyg-lib/pull/160))
 ### Changed
+- Added `fill_with_ints` function (it uses MKL library in order to generate ints) to RandintEngine class [#222](https://github.com/pyg-team/pyg-lib/pull/222)
```
rusty1s, line 21: Can we move this up to `0.3.0`?

rusty1s (approved):

rusty1s:
@OlhaBabicheva Can we look at the test failures?

rusty1s:
Looks like it cannot find `#include "mkl_vsl.h"`?

OlhaBabicheva:
> Looks like it cannot find `#include "mkl_vsl.h"`?

I changed it recently to #include <mkl.h> (I saw that we use the same include in  [matmul_kernel.cpp](https://github.com/pyg-team/pyg-lib/blob/72d2647f993a41565daee27cb3dcba6823dc200c/pyg_lib/csrc/ops/cpu/matmul_kernel.cpp#L18) ) ). However I'm still getting error during C++ testing (fatal error: mkl.h: No such file or directory)

rusty1s:
@DamianSzwichtenberg Do you have any advice here?

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -134,25 +134,50 @@ class NeighborSampler {
 
     // Case 2: Sample with replacement:
     else if (replace) {
-      for (size_t i = 0; i < count; ++i) {
-        const auto edge_id = generator(row_start, row_end);
-        add(edge_id, global_src_node, local_src_node, dst_mapper,
-            out_global_dst_nodes);
+      if (row_end < (1 << 16)) {
```
sdulloor, line 137: We could move this logic into `generator` instead of spilling it into the `sample` method.

OlhaBabicheva: Thanks for the review :)
However there is a problem with the type of the array. Because when we use RandintEngine we don't know, which type of number it'll return (int32 or int64). So I would have to assume, that we have the worst case scenario (int64), which won't be optimal space-wise. 
Moreover, in this case, I would have to create 2 arrays - the one, that will store numbers generated by mkl (unfortunately, mkl can't take pointers / references to int32/int64 type arrays) and the array that will handle cases of int32/int64. Thus, I'll still need to check, which one of the arrays was used in neighbor_kernel.cpp

DamianSzwichtenberg:
> @DamianSzwichtenberg Do you have any advice here?

@rusty1s @OlhaBabicheva Yes, I think this specific error occurs, because `test_rand_engine.cpp` includes `mkl` directly via nested includes. So the compiler does not know, where this file is. A possible solution for that could be:
```
if(MKL_INCLUDE_FOUND)
    target_include_directories(${name} PRIVATE ${BLAS_INCLUDE_DIR})
endif()
```
put [here](https://github.com/pyg-team/pyg-lib/blob/master/cmake/test.cmake#LL19C37-L19C49).

rusty1s:
Any updates on this one?

rusty1s:
Ok, let me know if I can help you with anything.

OlhaBabicheva:
@rusty1s Just wanted to let you know, that I fixed the building step (I had to change mkl to newer version in cpp_testing.yml), however there is one test ([test 19](https://github.com/pyg-team/pyg-lib/actions/runs/5431026827/jobs/9877141460?pr=222#step:6:379)) is failing

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -135,25 +135,50 @@ class NeighborSampler {
...
+        int arr[count];
+        generator.fill_with_ints(row_start, row_end, count, &arr[0]);
```
DamianSzwichtenberg, line 140: I think we should have an overloaded version of `operator()` specialized for mkl. That will help to have a single loop for both cases. Data for mkl case could be kept in `std::vector` on the `RandintEngine` side.

setup.py
```diff
@@ -93,17 +93,15 @@ def maybe_append_with_mkl(dependencies):
         torch_config = torch.__config__.show()
         with_mkl_blas = 'BLAS_INFO=mkl' in torch_config
         if torch.backends.mkl.is_available() and with_mkl_blas:
-            # product version is decoupled from library version. For older
-            # releases, where MKL was not a part of oneAPI, we can safely use
-            # 2021.4 as it is backward compatible (default for PyTorch conda
-            # distribution).
-            product_version = '2021.4'
+            product_version = '2023.1.0'
```
DamianSzwichtenberg, line 96: Could you please check whether this change affects `segment_matmul` performance?

DamianSzwichtenberg: Verified, no impact on `segment_matmul`.

pyg_lib/csrc/random/cpu/rand_engine.h
```diff
@@ -99,18 +103,46 @@ class PrefetchedRandint {
...
+    vslDeleteStream(&stream);
+#endif
+  }
 
   // Uniform random number within range [beg, end)
   T operator()(T beg, T end) {
     TORCH_CHECK(beg < end, "Randint engine illegal range");
 
     T range = end - beg;
+    if (!prefetch_initialized) {
```
DamianSzwichtenberg, line 122: Is this necessary? `prefetched_` is actually always initialized at this point.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -135,25 +135,50 @@ class NeighborSampler {
 
     // Case 2: Sample with replacement:
     else if (replace) {
-      for (size_t i = 0; i < count; ++i) {
-        const auto edge_id = generator(row_start, row_end);
-        add(edge_id, global_src_node, local_src_node, dst_mapper,
-            out_global_dst_nodes);
+      if (row_end < (1 << 16)) {
+        int arr[count];
```
DamianSzwichtenberg, line 139: `count` is not a constant expression, this will probably fail with some compilers.

pyg_lib/csrc/random/cpu/rand_engine.h
```diff
@@ -1,9 +1,13 @@
 #pragma once
 
 #include <ATen/ATen.h>
-
 #include <limits.h>
 
+#include "pyg_lib/csrc/config.h"
+#if WITH_MKL_BLAS()
```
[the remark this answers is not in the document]
DamianSzwichtenberg: Update: let's leave it for another, cleaning PR.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
DamianSzwichtenberg: Update: let's leave it for future enhancements.

pyg_lib/csrc/random/cpu/rand_engine.h
DamianSzwichtenberg: Fixed.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
DamianSzwichtenberg: Fixed.

DamianSzwichtenberg:
Cpp benchmark results:
```
----------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations
----------------------------------------------------------------------------------
BenchmarkRandEngineWithMKL/256                 186 ns          186 ns      3776603
BenchmarkRandEngineWithMKL/512                 374 ns          374 ns      1861909
BenchmarkRandEngineWithMKL/1024                600 ns          600 ns      1170031
BenchmarkRandEngineWithMKL/2048               1117 ns         1116 ns       626010
BenchmarkRandEngineWithMKL/4096               2199 ns         2199 ns       316950
BenchmarkRandEngineWithPrefetching/256        1972 ns         1972 ns       354124
BenchmarkRandEngineWithPrefetching/512        3951 ns         3951 ns       177682
BenchmarkRandEngineWithPrefetching/1024       7880 ns         7879 ns        88383
BenchmarkRandEngineWithPrefetching/2048      15779 ns        15779 ns        44113
BenchmarkRandEngineWithPrefetching/4096      31520 ns        31519 ns        22161
```

@OlhaBabicheva can you please re-do benchmarks from neighbor.py?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed performance issues reported by Coverity Tool ([#240](https://github.com/pyg-team/pyg-lib/pull/240))
 - Updated `cutlass` version for speed boosts in `segment_matmul` and `grouped_matmul` ([#235](https://github.com/pyg-team/pyg-lib/pull/235))
 - Drop nested tensor wrapper for `grouped_matmul` implementation ([#226](https://github.com/pyg-team/pyg-lib/pull/226))
+- Added `fill_with_ints` function (it uses MKL library in order to generate ints) to RandintEngine class ([#222](https://github.com/pyg-team/pyg-lib/pull/222))
```
DamianSzwichtenberg, line 16: @OlhaBabicheva Please update the function name (`generate_range_of_ints`).

DamianSzwichtenberg (approved):

PR #226, opened 2023-05-08
Drop nested tensor wrapper for `grouped_matmul`

puririshi98:
With recent pytorch updates, our original implementation with list of tensors should work, no need for nested tensors
There are still some cases where the for-loop + linear is faster than heterodictlinear but this varies with number of types and size of tensors, so i will learn another sklearn hueristic for HeteroDictLinear the same way i have done for  [HeteroLinear(segmatmul)](https://github.com/pyg-team/pytorch_geometric/pull/7258)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `metis` partitioning ([#229](https://github.com/pyg-team/pyg-lib/pull/229))
 - Enable `hetero_neighbor_samplee` to work in parallel ([#211](https://github.com/pyg-team/pyg-lib/pull/211))
 ### Changed
+- Drop nested tensor wrapper for `grouped_matmul` implementation ([#226](https://github.com/pyg-team/pyg-lib/pull/226))
 - Fixed TorchScript support in `grouped_matmul` ([#220](https://github.com/pyg-team/pyg-lib/pull/220))
 ### Removed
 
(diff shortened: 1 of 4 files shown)


rusty1s (commented):
This is pretty cool. Can you give some comments/documentation in code about `pytreeify` and why we need it since I am pretty unfamiliar with it.

puririshi98:
> This is pretty cool. Can you give some comments/documentation in code about `pytreeify` and why we need it since I am pretty unfamiliar with it.

Yeah ill defs add explaination comments once i get it working

rusty1s (approved):
Thank you. This is great. One thing I noticed is that the CUDA tests fails for me, and I am not sure why. Does it work for you?

puririshi98:
```
assert torch.allclose(outs[i], expected, atol=1e-6), str((outs[i]-expected).abs().max())
E           AssertionError: tensor(20.8059, device='cuda:0', grad_fn=<MaxBackward1>)
```
yeah very wierd, this is a huge error will investigate

puririshi98:
i tested that even master branch of pyg-lib, if I edit the python code to use our custom cpp/cuda kernel instead of nested tensor pathway cpu passes and cuda fails with the same magnitude of failure:
```
        for i in range(len(outs)):
            assert outs[i].size() == (inputs[i].size(0), others[i].size(-1))
>           assert torch.allclose(outs[i], inputs[i] @ others[i] + biases[i],
                                  atol=1e-6), str((outs[i] - inputs[i] @ others[i] + biases[i]).abs().max())
E           AssertionError: tensor(22.6574, device='cuda:0')
```
to me this implies the cuda kernel is the underlying issue, especially because cpu passes
I will loop in cutlass team

rusty1s:
Thanks for digging into this :) Pretty weird that `segment_matmul` works without a problem, so I expect that may be solely introduced by the different feature dimensionalities across inputs (maybe we are doing something wrong there).

puririshi98:
```
test/ops/test_matmul.py::test_segment_matmul_autograd PASSED
test/ops/test_matmul.py::test_grouped_matmul_autograd PASSED
```
works now

rusty1s:
Amazing :) Big thanks to you and the cutlass team!

PR #233, opened 2023-05-26
Draft: trying to use pad 4 to speed up things

puririshi98:


rusty1s:
Does this speed-up things? Do you have numbers? There is a trade-off here since padding introduces additional memory requirements. Another way to fix this would be to dynamically infer the right `pad_size`, and call the corresponding kernel based on that. This is probably the more preferred solution?

puririshi98:
checked its useless

PR #236, opened 2023-06-05
optmization for threadblock count suggested by CUTLASS team

puririshi98:


puririshi98:
for the normal ampere kernels it doesnt do much but it should provide speedboosts and is the correct protocol for other kernels for diff dtype or diff gpu compute capabilities

puririshi98:
from cutlass team:
"CUTLASS's grouped GEMM operates as a set of persistent threadblocks. In a "normal" GEMM launching non-persistent threadblocks, one typically launches as many threadblocks as there are tiles in the output of the GEMM, and each threadblock computes one tile. In persistent threadblocks, one instead launches a fixed set of threadblocks, each of which computes one or more output tiles.
For grouped GEMM, persistent threadblocks are beneficial because the CUTLASS grouped GEMM operator takes in the list of problem sizes as a device pointer and, thus, cannot determine the total number of output tiles across all problems in the group without some preprocessing. Thus, we launch a fixed set of threadblocks and each typically pulls multiple tiles from the output space across all problems in the group. There are other benefits of persistent threadblocks outside of the use of grouped GEMM, but those are beyond the scope of the current discussion.
From a physical resource perspective, it is preferable to launch the maximum number of threadblocks that can fit in a single "wave" on the SMs of our GPU. This is simply the SM count of the GPU multiplied by the number of threadblocks of our kernel that can reside on a single SM at the same time (i.e., the kernel's occupancy). The GemmGrouped::sufficient() method calculates this. Thus, it is preferable to use this method to calculate the number of threadblocks to launch in grouped GEMM."

rusty1s (approved):

PR #241, opened 2023-07-05
Implemented the first version of MessageQueue (using shared memory)

cfgfung:
This is the first version of the MessageQueue.

**Overall architecture**

image

**High-level code architecture**

image


rusty1s:
Any updates on this? Why did we close it?

mananshah99:
@cfgfung I think it would still be good to work together and merge this one in the future to help handle scenarios in which we have limited access to shared memory (or in the case that we have the potential to oversubscribe shared memory due to sampling being decoupled from data loading), even though the file_system sharing strategy helps avoid the deadlock with file_descriptor.

PR #242, opened 2023-07-09
An initial implementation of the LABOR-0 sampling algorithm

mfbalin:
Implements the LABOR-0 sampling algorithm described in https://arxiv.org/abs/2210.13339 and https://docs.dgl.ai/en/1.1.x/generated/dgl.dataloading.LaborSampler.html.
Sequential poisson sampling was used so that each vertex gets a deterministic number of neighbors fully matching the behaviour of neighbor sampler.

Update: To appear at NeurIPS 2023.


mfbalin:
I am not very familiar with disjoint, directed, or temporal options so for now deleted them. But LABOR-0 can be used as a drop-in replacement to neighbor sampling so it should be possible to extend its support to those cases too.

mfbalin:
@rusty1s What are the next steps to seamlessly enable users to experiment with this new sampler in PyG?

mfbalin:
Right now, only LABOR-0 is implemented. For the importance sampling versions, an edge vector containing reciprocals of importance sampling probabilities will be returned. Should I add that vector to the return value as an optional return value before that case is implemented?

mfbalin:
If weighted sampling is of interest to PyG users, I can contribute that too.

mfbalin:
[labor2023-07-13 034625.678968.csv](https://github.com/pyg-team/pyg-lib/files/12035180/labor2023-07-13.034625.678968.csv)

mfbalin:
Is there an easy way to refactor the neighbor_kernel and this file so that the extra features such as temporal or disjoint come for free for this implementation as well?

mfbalin:
CPU and GPU implementations that support weighted sampling can be found here: https://docs.dgl.ai/en/latest/generated/dgl.graphbolt.LayerNeighborSampler.html#dgl.graphbolt.LayerNeighborSampler
It is possible to use this library as a dataloader for PyG models so there is not much point to maintain another version here.

PR #246, opened 2023-08-21
Add support for distributed sampling

kgajdamo:
This code belongs to the part of the whole distributed training for PyG.

## Description
Distributed training neighbor sampling differs from the sampling currently implemented in pyg-lib. During distributed training nodes from one batch can be sampled by different machines (and therefore different samplers). The result of this is incorrect subtree/subgraph node indexing.
To achieve correct results it is necessary to sample by one hop and then synchronise outputs between machines.

Proposed algorithm:
1. First sample only global node ids (`sampled_nodes`) with duplicates in `neighbor_sample`.
2. Do not sample rows and cols but save information of how many neighbors were sampled by each node (`cumm_sum_sampled_nbrs_per_node`).
3. After each layer: synchronise and merge outputs from different machines and take new seed nodes (without duplicates) from sampled_nodes.
4. Sample next layer and continue 1-3 until all layers are sampled.
5. Perform global to local mappings using mapper and create (row, col) based on a `sampled_nodes_with_duplicates` and  `sampled_nbrs_per_node`.

Step 3. was implemented in pytorch_geometric.

## Added
- new argument `distributed` to the `neighbor_sample` function to enable the algorithm described above.
- new argument `batch` to the `neighbor_sample` function that allows to specify the initial subgraph indices for seed nodes (used with disjoint).
- new return value `cumm_sum_sampled_nbrs_per_node` to the `neighbor_sample` function to return cumulative sum of the sampled neighbors per each node.

- new function `relabel_neighborhood` that is used after sampling all layers and its purpose is to relabel global indices of the sampled nodes to the local subtree/subgraph indices (row, col).
- new function `hetero_relabel_neighborhood` (same as `relabel_neighborhood` but for heterogeneous graphs). Returns (row_dict and col_dict).
- unit tests


pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
```diff
@@ -0,0 +1,261 @@
...
+                "Batch needs to be specified to create disjoint subgraphs");
+    TORCH_CHECK(batch.value().is_contiguous(), "Non-contiguous 'batch'");
+    TORCH_CHECK(batch.value().numel() == sampled_nodes_with_dupl.numel(),
+                "Each node must belong to a subgraph.'");
+  }
+  TORCH_CHECK(seed.is_contiguous(), "Non-contiguous 'seed'");
+  TORCH_CHECK(sampled_nodes_with_dupl.is_contiguous(),
+              "Non-contiguous 'sampled_nodes_with_dupl'");
+
+  at::Tensor out_row, out_col;
```
DamianSzwichtenberg, line 50: I believe creating an empty vector is cheaper than creating a Tensor. You could declare here `sampled_rows` and `sampled_cols` and at the end simply return `get_sampled_edges<scalar_t>(sampled_rows, sampled_cols, csc);`

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+template <bool disjoint>
+std::tuple<c10::Dict<rel_type, at::Tensor>, c10::Dict<rel_type, at::Tensor>>
+relabel(const std::vector<node_type>& node_types,
+        const std::vector<edge_type>& edge_types,
+        const c10::Dict<node_type, at::Tensor>& seed_dict,
+        const c10::Dict<node_type, at::Tensor>& sampled_nodes_with_dupl_dict,
+        const c10::Dict<node_type, std::vector<int64_t>>&
+            sampled_nbrs_per_node_dict,
+        const c10::Dict<node_type, int64_t> num_nodes_dict,
```
DamianSzwichtenberg, line 106: Missing `&` in type?

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
```diff
@@ -0,0 +1,261 @@
...
+        const c10::optional<c10::Dict<node_type, at::Tensor>>& batch_dict,
+        const bool csc) {
+  if (disjoint) {
```
DamianSzwichtenberg, line 109: `if constexpr`?

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
```diff
@@ -0,0 +1,261 @@
...
+    TORCH_CHECK(batch_dict.has_value(),
+                "Batch needs to be specified to create disjoint subgraphs");
+    for (const auto& kv : batch_dict.value()) {
+      const at::Tensor& batch = kv.value();
```
DamianSzwichtenberg, line 113: Shouldn't these checks be performed at `dist_neighbor.cpp`?
BTW, is this intended to have `batch` and `sampled_nodes_with_dupl` point to the same value? If so, does the last check make sense?

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
```diff
@@ -0,0 +1,261 @@
...
+                  "Each node must belong to a subgraph.'");
+    }
+  }
+  for (const auto& kv : seed_dict) {
+    const at::Tensor& seed = kv.value();
+    TORCH_CHECK(seed.is_contiguous(), "Non-contiguous 'seed'");
+  }
+  for (const auto& kv : sampled_nodes_with_dupl_dict) {
+    const at::Tensor& sampled_nodes_with_dupl = kv.value();
+    TORCH_CHECK(sampled_nodes_with_dupl.is_contiguous(),
```
DamianSzwichtenberg, line 126: Same as above, can this check be moved to `dist_neighbor.cpp`?

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
```diff
@@ -0,0 +1,261 @@
...
+        }
+        for (const auto& k : edge_types) {
+          const auto src = !csc ? std::get<0>(k) : std::get<2>(k);
+          const auto dst = !csc ? std::get<2>(k) : std::get<0>(k);
+          for (auto i = 0; i < sampled_nbrs_per_node_dict.at(dst).size(); i++) {
+            auto& dst_mapper = mapper_dict.at(dst);
+            auto& dst_sampled_nodes_data = sampled_nodes_data_dict.at(dst);
+            slice_dict.at(dst).second += sampled_nbrs_per_node_dict.at(dst)[i];
+            size_t begin, end;
+            std::tie(begin, end) = slice_dict.at(dst);
```
DamianSzwichtenberg, line 184: Maybe change to `auto [begin, end] = slice_dict.at(dst);`?

pyg_lib/csrc/sampler/dist_neighbor.cpp
```diff
@@ -0,0 +1,86 @@
...
+std::tuple<at::Tensor, at::Tensor> relabel_neighborhood(
+    const at::Tensor& seed,
+    const at::Tensor& sampled_nodes_with_dupl,
+    const std::vector<int64_t>& sampled_nbrs_per_node,
+    const int64_t num_nodes,
+    const c10::optional<at::Tensor>& batch,
+    bool csc,
+    bool disjoint) {
+  at::TensorArg seed_t{seed, "seed", 1};
+  at::TensorArg sampled_nodes_with_dupl_t{sampled_nodes_with_dupl,
```
DamianSzwichtenberg, line 20: This name suggests it is an alias to a type. A better option would be `_v` or `_arg` as a suffix.

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
```diff
@@ -0,0 +1,261 @@
...
+#include "pyg_lib/csrc/utils/types.h"
+
+namespace pyg {
+namespace sampler {
+
+namespace {
+
+template <typename scalar_t>
+std::tuple<at::Tensor, at::Tensor> get_sampled_edges(
+    std::vector<scalar_t> sampled_rows,
```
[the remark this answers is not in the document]
kgajdamo: Agree.

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
kgajdamo: scalar_t will not work outside of the macro

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
kgajdamo: My mistake. They should not point to the same value.

pyg_lib/csrc/sampler/cpu/dist_neighbor_kernel.cpp
kgajdamo: Moved.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -216,19 +226,25 @@ class NeighborSampler {
...
            std::vector<int64_t>,
+           std::vector<int64_t>,
            std::vector<int64_t>>
 sample(const at::Tensor& rowptr,
        const at::Tensor& col,
        const at::Tensor& seed,
        const std::vector<int64_t>& num_neighbors,
        const c10::optional<at::Tensor>& time,
        const c10::optional<at::Tensor>& seed_time,
+       const c10::optional<at::Tensor>& batch,
```
rusty1s, line 247: How does `batch` behave in case of `disjoint=False`? If distributed sampling requires `disjoint=True` anyway, I am not totally sure I understand why we need this new argument here.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -637,21 +706,25 @@ std::tuple<at::Tensor,
            at::Tensor,
            c10::optional<at::Tensor>,
            std::vector<int64_t>,
+           std::vector<int64_t>,
            std::vector<int64_t>>
 neighbor_sample_kernel(const at::Tensor& rowptr,
```
rusty1s, line 711: Are you okay with adding `distributed_neighbor_sample` function instead at a high-level? This way, we don't necessarily need to make changes to the interface for `distributed=False`. We can still use `DISPATCH_SAMPLE` internally in these functions.

rusty1s (commented):
Thanks for the PR. A few minor high-level comments about the neighbor sampling part.

Overall, I think it would be easier to get this PR in if we would add `merge_sampler_outputs` and `relabel_neighborhood` separately.

kgajdamo:
> Thanks for the PR. A few minor high-level comments about the neighbor sampling part.
> 
> Overall, I think it would be easier to get this PR in if we would add `merge_sampler_outputs` and `relabel_neighborhood` separately.

Thank you for the comments. As you suggested I opened a new PR for `merge_sampler_outputs`: [#252](https://github.com/pyg-team/pyg-lib/pull/252 )

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -301,6 +328,8 @@ sample(const at::Tensor& rowptr,
           sampler.uniform_sample(/*global_src_node=*/sampled_nodes[i],
                                  /*local_src_node=*/i, count, mapper, generator,
                                  /*out_global_dst_nodes=*/sampled_nodes);
+          if constexpr (distributed)
+            cumm_sum_sampled_nbrs_per_node.push_back(sampled_nodes.size());
```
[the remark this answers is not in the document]
kgajdamo: `num_sampled_nodes_per_hop` and `cmm_sum_sampled_nbrs_per_node` are quite different. `num_sampled_nodes_per_hop` refers to the number of all nodes sampled in a layer, however `cmm_sum_sampled_nbrs_per_node` saves information of how many neighbors were sampled by each src node. So for example if there will be seed size = 4, and num_neighbors=[2], then `num_sampled_nodes_per_hop ` = [4, 8] and `cmm_sum_sampled_nbrs_per_node`=[4, 6, 8, 10, 12]. Based on the `cmm_sum_sampled_nbrs_per_node` we could calculate `num_sampled_nodes_per_hop`. Otherwise, I don't think so.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
kgajdamo: Distributed sampling does not requires `disjoint=true`. It can work with `disjoint=false` as well.
`batch` is used only when `disjoint=true`, otherwise it is not relevant.
Why `batch` is needed:
During distributed sampling we sample by one hop in c++ and go out of the sample() function. So if we sample more than one layer, information about which subgraph a given node belonged to, will be lost. So, thanks to the `batch` variable, we can assign initial values.

rusty1s (approved):

rusty1s:
Can we also add a test for these functions?

kgajdamo:
> Thanks @kgajdamo. To decrease the size of PR, I removed the `relabel` function. Can you recheck it in in a separate PR - please be patient with me :(
> 
> In this PR, I made the following changes:
> 
> * I added `dist_neighbor_sample` and `dist_hetero_neighbor_sample`. Please call these on PyG side since I don't want to make the general `neighbor_sample` interface to polluted. For `dist_neighbor_sample` we can also consider cleaning up the interface, e.g., `num_neighbors` should only be an `int` rather than a `list`, but it's not a must.
> * For now, I removed the `batch` argument since I still don't see why we need it. Please bring it back if you see no other way to resolve this.

Thanks for the updates. It is a good idea to have a `dist_neighbor_sample` function instead of using the `neighbor_sample`. Here are some additional comments from me:
* due to the fact, that distributed sampling have a loop over the layers in python, in case of hetero at the moment when we call `neighbor_sample` we have only one edge type. So it becomes actually homo and we don't need the `dist_hetero_neighbor_sample` because we can use `dist_neighbor_sample` instead. I removed then `dist_hetero_neighbor_sample` function. 
* I also removed all not used outputs and left only: `node`, `edge_ids`, `cummsum_sampled_nbrs_per_node`. So it is more clear now.
* and changed `std::vector<int64_t> num_neighbors` input list into `int64_t one_hop_num` as you suggested.
* here is the new PR with the above mentioned changes and new unit tests (without relabel function, which I will add in the next one): [#253](https://github.com/pyg-team/pyg-lib/pull/253)

PR #253, opened 2023-09-06
Update `dist_neighbor_sample`

kgajdamo:
This code belongs to the part of the whole distributed training for PyG.

This PR is complementary to the [#246](https://github.com/pyg-team/pyg-lib/pull/246) and introduces some updates.

What has been changed:
* Removed not needed `dist_hetero_neighbor_sample` function (due to the fact, that distributed sampling have a loop over the layers in python, in case of hetero at the moment when we call `neighbor_sample` we have only one edge type. So it becomes actually homo and we don't need the `dist_hetero_neighbor_sample` and can use `dist_neighbor_sample` instead.)
* Removed all not used outputs and left only the following: `node`, `edge_ids`, `cummsum_sampled_nbrs_per_node`.
* Changed `std::vector<int64_t> num_neighbors` input list into `int64_t one_hop_num`.

Added:
* Unit tests

### DIFF
--- a/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
+++ b/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
@@ -847,17 +847,11 @@ hetero_neighbor_sample_kernel(
                   edge_weight_dict, csc, temporal_strategy);
 }
 
-std::tuple<at::Tensor,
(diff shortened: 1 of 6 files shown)


rusty1s (approved):
Thank you. Like it :)

PR #264, opened 2023-10-12
Add `softmax_csr` implementation

DamianSzwichtenberg:
This PR adds forward and backward implementation of sparse softmax operation as defined [here](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/utils/softmax.py#L9).

In the `pytorch_geometric` implementation we cannot take advantage of model compilation when groups are defined via `ptr`. `softmax_csr` introduced here provides a well-performing kernel for such a scenario.

Performance boost (achieved on 28C, single socket machine):
~7x for forward pass
~8x for backward pass
Additionally, GAT training time was reduced by ~5%.

### DIFF
--- a/pyg_lib/csrc/ops/cpu/softmax_kernel.cpp
+++ b/pyg_lib/csrc/ops/cpu/softmax_kernel.cpp
@@ -127,10 +127,44 @@ at::Tensor softmax_forward_kernel(const at::Tensor& src,
 
 at::Tensor softmax_backward_kernel_ptr_dim0_impl(const at::Tensor& out,
                                                  const at::Tensor& out_grad,
-                                                 const at::Tensor& ptr) {
+                                                 const at::Tensor& groups) {
   auto in_grad = at::zeros_like(out);
 
-  // TODO: not implemented yet
+  AT_DISPATCH_FLOATING_TYPES(
+      out.scalar_type(), "softmax_backward_kernel_ptr_dim0_impl", [&] {
+        const auto n_groups = groups.size(0) - 1;
+        const auto n_heads = out.size(-1);
+
+        const auto out_ptr = out.data_ptr<scalar_t>();
+        const auto out_grad_ptr = out_grad.data_ptr<scalar_t>();
+        const auto groups_ptr = groups.data_ptr<int64_t>();
+        auto in_grad_ptr = in_grad.data_ptr<scalar_t>();
+        const auto new_groups = std::move(
+            create_per_thread_groups(groups_ptr, n_groups, out.size(0)));
+
+        at::parallel_for(
+            0, new_groups.size() - 1, 1, [&](int64_t beg, int64_t end) {
+              for (auto group_id = new_groups[beg]; group_id < new_groups[end];
+                   ++group_id) {
+                const auto row_beg = groups_ptr[group_id];
+                const auto row_end = groups_ptr[group_id + 1];
+                const auto rows_in_group = row_end - row_beg;
+                const auto offset = row_beg * n_heads;
+                const auto out_beg_ptr = out_ptr + offset;
+                const auto out_grad_beg_ptr = out_grad_ptr + offset;
+                auto in_grad_beg_ptr = in_grad_ptr + offset;
+
+                for (int64_t i = 0; i < rows_in_group; ++i) {
+                  for (int64_t j = 0; j < rows_in_group * n_heads; ++j) {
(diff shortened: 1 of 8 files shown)


DamianSzwichtenberg:
@pyg-team/intel-team Please take a look.

pyg_lib/ops/__init__.py
```diff
@@ -332,6 +332,84 @@ def index_sort(
...
+
+        >>> src = torch.randn(4, 4)
+        >>> ptr = torch.tensor([0, 4])
+        >>> softmax(src, None, ptr)
+        tensor([[0.0157, 0.0984, 0.1250, 0.4523],
+                [0.1453, 0.2591, 0.5907, 0.2410],
+                [0.0598, 0.2923, 0.1206, 0.0921],
+                [0.7792, 0.3502, 0.1638, 0.2145]])
+    """
+    if src.dim() != 2 or not src.is_cpu or ptr is None or dim != 0:
```
kgajdamo, line 403: I wonder why ptr is optional because if you don't provide it, you get an error.

DamianSzwichtenberg: I wanted to make API in its final form, otherwise, each change here would require a change in `pytorch_geometric`. I'll add support for `index` in the near future.

DamianSzwichtenberg: @kgajdamo, after rethinking your suggestion, I decided to change the API and create a specialized `softmax_csr` operation that accepts `ptr` only. Rationale:
`torch.compile` gives nice results for `softmax` with groups defined via `index`, hence I don't see a reason to have a specialized kernel for that option.

DamianSzwichtenberg:
@kgajdamo @rusty1s Please take a look. I made the softmax implementation a bit more general, so now it covers any `src` dimensionality as well as any given `dim`. I also restricted groups to be defined via `ptr` ([rationale here](https://github.com/pyg-team/pyg-lib/pull/264#discussion_r1381658840)).

yanbing-j:
Hi @DamianSzwichtenberg , this PR looks good to me. The overall structure of softmax kernel with sparse input is similar with that in softmax kernel of dense input in PyTorch. With the sparsity, the performance boost is from parallelism, right?

And will this PR upstream to PyTorch later? Since there is no SparseCsr support for softmax yet in PyTorch.

pyg_lib/csrc/ops/cpu/softmax_kernel.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+at::Tensor softmax_csr_forward_kernel_impl(const at::Tensor& src,
+                                           const at::Tensor& groups,
+                                           const int64_t dim) {
+  auto out = at::zeros_like(src);
+
+  AT_DISPATCH_FLOATING_TYPES(
+      src.scalar_type(), "softmax_csr_forward_kernel_impl", [&] {
+        const auto n_groups = groups.size(0) - 1;
+        const auto n_heads = src.numel() / src.size(dim);
```
kgajdamo, line 63: I assume that n_heads should be of integer type, so src.numel() will always be divisible by src.size(dim)? Maybe I'd worth to add assert about that. Same for the backward.

DamianSzwichtenberg: I think it should be guaranteed. `numel()` can be represented as:
```python
numel_ = 1
for i in range(src.dim()):
    numel_ *= src.size(i)
```
As `numel` represents product of  `size`'s, it is divisible by them.

kgajdamo (approved):
Looks good to me.

pyg_lib/csrc/ops/cpu/softmax_kernel.cpp
kgajdamo: Ok :)

pyg_lib/ops/__init__.py
```diff
@@ -332,6 +331,62 @@ def index_sort(
     return torch.ops.pyg.index_sort(inputs, max_value)
 
 
+class Softmax(torch.autograd.Function):
+    @staticmethod
+    def forward(
```
rusty1s, line 336: Can we define the autograd function directly in C++?

DamianSzwichtenberg: Should be possible, will check.

DamianSzwichtenberg: #282

PR #270, opened 2023-10-31
Optimization of biased sampling

OlhaBabicheva:
Implemented biased sampling using this algorithm: https://utopia.duth.gr/~pefraimi/research/data/2007EncOfAlg.pdf
image

### DIFF
--- a/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
+++ b/pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
@@ -224,12 +224,14 @@ class NeighborSampler {
     // Case 2: Multinomial sampling:
     else {
       at::Tensor index;
-      if (replace) {  // at::multinomial has good perfomance only when
-                      // replace=true, e.g.
-                      // https://github.com/pytorch/pytorch/issues/11931#top
+      if (replace) {
+        // at::multinomial only has good perfomance for `replace=true`, see:
+        // https://github.com/pytorch/pytorch/issues/11931
         index = at::multinomial(weight, count, replace);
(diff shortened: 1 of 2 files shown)


pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -223,11 +223,22 @@ class NeighborSampler {
...
-      for (size_t i = 0; i < index.numel(); ++i) {
-        add(row_start + index_data[i], global_src_node, local_src_node,
-            dst_mapper, out_global_dst_nodes);
+      if (replace) {
+        const auto index = at::multinomial(weight, count, replace);
+        const auto index_data = index.data_ptr<int64_t>();
+        for (size_t i = 0; i < index.numel(); ++i) {
+          add(row_start + index_data[i], global_src_node, local_src_node,
+              dst_mapper, out_global_dst_nodes);
+        }
```
rusty1s, line 232: I think we can unify this logic outside the `if`/`else` to avoid the duplicated code here.

pyg_lib/csrc/sampler/cpu/neighbor_kernel.cpp
```diff
@@ -223,11 +223,22 @@ class NeighborSampler {
...
+      } else {
+        const auto rand = at::empty_like(weight).uniform_();
+        const auto key = (rand.log() / weight);
```
rusty1s, line 235: Can you add a pointer here to the algorithm, the PyTorch issue, and why it is correct? :)

rusty1s (approved):

PR #282, opened 2023-11-20
Move `softmax_csr` autograd module to C++

DamianSzwichtenberg:


pyg_lib/ops/__init__.py
```diff
@@ -384,7 +361,7 @@ def softmax_csr(
                 [0.7792, 0.3502, 0.1638, 0.2145]])
     """
     dim = dim + src.dim() if dim < 0 else dim
-    return Softmax.apply(src, ptr, dim)
+    return torch.ops.pyg.softmax_csr_forward(src, ptr, dim)
```
rusty1s, line 364:
```suggestion
    return torch.ops.pyg.softmax_csr(src, ptr, dim)
```

rusty1s (approved):

pyg_lib/ops/__init__.py
DamianSzwichtenberg: This requires changes in C++ code as well, but I agree that `softmax_csr_forward` is a bit misleading here.

PR #308, opened 2024-03-14
fix bug to add large tensor support

kaixuanliu:


kaixuanliu:
I ran into an error as I define a very large sparse tensor in the following code,
`
import torch
import torch_sparse

num_edges = 7982968874
num_src_nodes = 264388606
edge_id = torch.arange(num_edges, dtype=torch.int64)
node_sizes = (num_src_nodes + 1, num_src_nodes + 1)
row = torch.randint(num_src_nodes, (num_edges, ), dtype=torch.int64)
col = torch.randint(num_src_nodes, (num_edges, ), dtype=torch.int64)
adj_t = torch_sparse.SparseTensor(
            row=row, col=col, value=edge_id, sparse_sizes=node_sizes
        ) 
`
It will crash in the former implementation due to incorrect data type conversion, this PR solved this kind of data overflow

rusty1s (approved):

rusty1s:
Thank you:)

PR #415, opened 2025-03-19
Implement a basic hetero sampler in the class

vid-koci:
Implement a basic GraphSAGE sampler inside the `HeteroNeighborSampler`.


pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
          bool disjoint = false,
          std::string temporal_strategy = "uniform",
          bool return_edge_id = true) {
-    // TODO
-    c10::Dict<rel_type, at::Tensor> row;
-    c10::Dict<rel_type, at::Tensor> col;
-    c10::Dict<node_type, at::Tensor> node_id;
-    c10::Dict<rel_type, at::Tensor> edge_id;
+    if (edge_time_.has_value())
+      TORCH_CHECK(seed_time.has_value(), "Seed time needs to be specified");
```
rusty1s, line 182: and `node_time_`? Perhaps we can add a utility function such as `is_temporal`.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+      }
+    }
+    TORCH_CHECK(temporal_strategy == "uniform" || temporal_strategy == "last",
+                "No valid temporal strategy found");
+    clear_placeholders();
+
+    pyg::random::RandintEngine<int64_t> generator;
+
+    for (const auto& k : edge_types_) {
+    }
```
rusty1s, line 196: ?

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+    phmap::flat_hash_map<node_type, size_t> num_nodes_dict;
+    for (const auto& k : edge_types_) {
+      const auto num_nodes = rowptr_.at(to_rel_type(k)).size(0) - 1;
+      num_nodes_dict[std::get<0>(k)] = num_nodes;
+    }
```
rusty1s, line 201: Why don't we do it in init?

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+    // Add node types that only exist in the seed_node
+    // TODO why would this happen?
+    for (const auto& kv : seed_node) {
+      const at::Tensor& seed = kv.value();
+      if (num_nodes_dict.count(kv.key()) == 0 && seed.numel() > 0) {
+        num_nodes_dict[kv.key()] = seed.max().data_ptr<int64_t>()[0] + 1;
+      }
+    }
```
rusty1s, line 209: is this copy-pasted? Don't get it yet. I assume it's a fallback if we cannot infer full `num_nodes_dict` from `rowptr` alone. I would be also fine to just pass `num_nodes_dict` to `NeighborSampler.init` rather than relying on this hack here.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+
+    // Important: Since `disjoint` isn't known at compile time, we just use
+    // pairs <int64, int64> for both usecases and fill batch with 0
+    // for !disjoint case. This is a bit wasteful but we'll clean this up
+    // sometime later. TODO
```
rusty1s, line 214: You can just make it a template arg. Not urgent.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+      TORCH_CHECK(seed_time.has_value(), "Seed time needs to be specified");
+    if (seed_time.has_value()) {
+      for (const auto& kv : seed_time.value()) {
+        const at::Tensor& seed_time = kv.value();
+        TORCH_CHECK(seed_time.is_contiguous(), "Non-contiguous 'seed_time'");
+      }
+    }
+    TORCH_CHECK(temporal_strategy == "uniform" || temporal_strategy == "last",
+                "No valid temporal strategy found");
+    clear_placeholders();
```
rusty1s, line 191: I would assume we clear intermediate data at the end, not at the start.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+    const bool parallel = at::get_num_threads() > 1 && edge_types_.size() > 1;
+    std::vector<std::vector<edge_type>> threads_edge_types;
+
+    size_t L = 0;  // num_layers
+    // Split edge types into threads
+    for (const auto& k : edge_types_) {
+      num_sampled_edges_per_hop_.insert(
+          {to_rel_type(k), std::vector<int64_t>(1, 0)});
+      L = std::max(L, num_neighbors.at(to_rel_type(k)).size());
+      if (parallel) {
```
rusty1s, line 241: I am not a big fan of the multi-threading logic here (or better speaking in general) because it complicates code logic quite a bit. During data loading, we only have a single thread in the first place (so `parallel` will always be `False`), since we parallelize sampling across batches (via `num_workers>0`). Multi-threading here is only helpful if we run with `num_workers=0`, which no one really does.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+                auto& src_sampled_nodes = sampled_nodes.at(src);
+                auto& dst_sampled_nodes = parallel
+                                              ? dst_sampled_nodes_dict.at(dst)
+                                              : sampled_nodes.at(dst);
+                auto& dst_mapper = mapper_dict.at(dst);
+                size_t begin, end;
+                std::tie(begin, end) = slice_dict.at(src);
+                num_sampled_edges_per_hop_[to_rel_type(k)].push_back(0);
+                // We skip weighted/biased edges and edge-tempora sampling for
+                // now If no timestamps are involved
```
rusty1s, line 339:
```suggestion
                // We skip weighted/biased edges and edge-temporal sampling for
                // now
```

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -85,26 +178,367 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+          index_tracker.insert(i);
+        }
+        const auto edge_id = row_start + rnd;
+        add(e_type, edge_id, global_src_node, local_src_node, dst_mapper,
+            out_global_dst_nodes, return_edge_id);
+      }
+    }
+  }
+
+  inline void add(rel_type e_type,
```
rusty1s, line 499: can we come up with descriptive name?

rusty1s (approved):
Great start.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -1,6 +1,12 @@
 #include <ATen/ATen.h>
+#include <ATen/Parallel.h>
 #include <torch/library.h>
 
+#include "pyg_lib/csrc/random/cpu/rand_engine.h"
+#include "pyg_lib/csrc/sampler/cpu/index_tracker.h"
+#include "pyg_lib/csrc/sampler/cpu/mapper.h"
+#include "pyg_lib/csrc/sampler/cpu/neighbor_kernel.h"
```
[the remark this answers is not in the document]
vid-koci: I believe this was a leftover from testing - thanks for catching

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -56,21 +62,108 @@ struct NeighborSampler : torch::CustomClassHolder {
...
+      }
+    }
+    if (edge_time.has_value()) {
+      for (const auto& kv : edge_time.value()) {
+        const at::Tensor& time = kv.value();
+        TORCH_CHECK(time.is_contiguous(), "Non-contiguous 'edge_time'");
+      }
+    }
+    for (const auto& k : edge_types_)
+      phmap::flat_hash_map<node_type, size_t> num_nodes_dict;
```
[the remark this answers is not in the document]
vid-koci: Oops, looks like a leftover from debugging

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: The original implementation logic which I was following will just use node times from the graph if seed_time isn't given. I don't see a reason to deviate from the behaviour of `hetero_neighbor_sample`

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: I really should've been more careful when removing the debug statements.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: Yeah, this is copy-pasted. I'll move it to the constructor together with the logic above it.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: See my answer to the next comment

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: Agree - let's do this as part of the proper public-facing implementation.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: Thanks, that's reasonable!

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: I removed parallel processing. If there's little benefit to it then I see no reason why to maintain it.

PR #417, opened 2025-03-24
Add `MetapathTracker` helper class to `NeighborSampler`

vid-koci:
Helper class that keeps track of all metapaths and their expected number in the sample.
Only expected for internal use.


pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -60,6 +60,112 @@ struct NeighborSampler : torch::CustomClassHolder {
   const c10::optional<at::Tensor>& edge_time_;
 };
 
+struct MetapathTracker : torch::CustomClassHolder {
```
rusty1s, line 63: Can we add some notes/comments here on what this does?

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -557,6 +663,15 @@ TORCH_LIBRARY_FRAGMENT(pyg, m) {
...
       .def("sample", &HeteroNeighborSampler::sample);
+
+  m.class_<MetapathTracker>("MetapathTracker")
+      .def(torch::init<std::vector<edge_type>,
+                       c10::Dict<rel_type, std::vector<int64_t>>,
+                       std::vector<node_type>>())
+      .def("report_sample_size", &MetapathTracker::report_sample_size)
+      .def("get_sample_size", &MetapathTracker::get_sample_size)
+      .def("init_batch", &MetapathTracker::init_batch)
+      .def("get_neighbor_metapath", &MetapathTracker::get_neighbor_metapath);
```
rusty1s, line 674: why do you need to expose? If this is just for testing purposes, let's add a C++ test instead.

rusty1s (approved):

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: You and your perfectly reasonable and sensible suggestions...

I've done it

PR #419, opened 2025-03-25
Oversample metapaths where number of samples was lower than expected

vid-koci:
Use `MetapathTracker` to compute the expected number of nodes of a type and adjust sampling as required.
Unfortunately, that requires shuffling nodes and thus loses determinism of temporal sampling.


pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -441,40 +470,98 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+                  if (batch_total_count.find(batch) ==
+                      batch_total_count.end()) {
+                    batch_total_count[batch] = 0;
+                    batch_processed_count[batch] = 0;
+                  }
+                  batch_total_count[batch]++;
+                }
+                // Whenever we undersample nodes in a batch due to the lack of
+                // neighbors, we allow oversampling neighbors of later nodes
+                // later nodes of the same batch. Later nodes could thus get
```
rusty1s, line 488:
```suggestion
                // neighbors, we allow oversampling neighbors of later nodes
                // of the same batch. Later nodes could thus get
```

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -441,40 +470,98 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+                // a larger number of neighbors. To avoid bias from the edge
+                // order, we check nodes in the random order.
+                std::vector<size_t> node_permutation;
+                for (size_t i = begin; i < end; ++i)
+                  node_permutation.push_back(i);
+                // TODO: replace with something based on `generator`
+                std::shuffle(node_permutation.begin(), node_permutation.end(),
+                             std::mt19937{std::random_device{}()});
```
rusty1s, line 496: Why not just do
```
perm = at::randperm(end - begin);
perm_data = perm.data_ptr<int64_t>();

File: test/classes/test_neighbor_sampler.py [hunk 3]
```diff
@@ -50,26 +50,39 @@ def test_hetero_neighbor_sampler_temporal_sample() -> None:
...
-    assert torch.equal(col['B__to__A'], torch.tensor([2, 3, 4]))
-    assert torch.equal(edge_id['A__to__B'], torch.tensor([3, 0]))
-    assert torch.equal(edge_id['B__to__A'], torch.tensor([1, 0, 1]))
-    assert torch.equal(node_id['A'], torch.tensor([2, 0, 1, 0, 1]))
-    assert torch.equal(node_id['B'], torch.tensor([1, 0, 1]))
-    assert torch.equal(batch['A'], torch.tensor([0, 1, 2, 0, 1]))
-    assert torch.equal(batch['B'], torch.tensor([2, 0, 1]))
-    assert num_sampled_nodes == {'A': [2, 1, 2], 'B': [1, 2, 0]}
-    assert num_sampled_edges == {'A__to__B': [0, 2, 0], 'B__to__A': [0, 1, 2]}
+    # Due to random shuffle, the output isn't entirely deterministic
```
rusty1s, line 61: if you set a seed here, and replace the `shuffle` logic, it should be.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: Thanks, exactly what I needed!

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -217,33 +231,36 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
   };
 
   void uniform_sample(rel_type e_type,
-                      const pair_int64_t global_src_node,
+                      const triple_int64_t global_src_node,
```
rusty1s, line 234: why do we need to pass a triplet here. Can we instead just introduce 3 arguments? Alternatively, can we document input/output somewhere?

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -314,10 +332,11 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
     }
 
     // Important: Since `disjoint` isn't known at compile time, we just use
-    // pairs <int64, int64> for both usecases and fill batch with 0
+    // tuples <int64, int64, int64> for both usecases and fill batch with 0
     // for !disjoint case. This is a bit wasteful but we'll clean this up
     // sometime later. TODO
-    phmap::flat_hash_map<node_type, std::vector<pair_int64_t>> sampled_nodes;
+    // Each triple is (batch_id, node_id, metapath_id)
```
rusty1s, line 338: triple or triplet?

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -386,59 +415,109 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+            batch_total_count[batch] = 0;
+            batch_processed_count[batch] = 0;
+          }
+          batch_total_count[batch]++;
+        }
+        // Whenever we undersample nodes in a batch due to the lack of
+        // neighbors, we allow oversampling neighbors of later nodes
+        // of the same batch. Later nodes could thus get
+        // a larger number of neighbors. To avoid bias from the edge
+        // order, we check nodes in the random order.
```
rusty1s, line 450: it's not exactly clear to me yet why we need to shuffle.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -386,59 +415,109 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+        auto perm = at::randperm(end - begin);
+        auto node_data = perm.data_ptr<int64_t>();
+        std::vector<int64_t> node_permutation;
+        for (int i = 0; i < end - begin; i++) {
+          node_permutation.push_back(node_data[i] + begin);
+        }
```
rusty1s, line 456: nit: Why do we need to convert to `std::vector` here? Can't we just iterate over `node_data` later on?

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -386,59 +415,109 @@ struct HeteroNeighborSampler : torch::CustomClassHolder {
...
+        // We skip weighted/biased edges and edge-tempora sampling for
+        // now If no timestamps are involved
```
rusty1s, line 458:
```suggestion
        // now If no timestamps are involved
```
weird comment.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
```diff
@@ -615,6 +700,15 @@ TORCH_LIBRARY_FRAGMENT(pyg, m) {
...
       .def("sample", &HeteroNeighborSampler::sample);
+
+  m.class_<MetapathTracker>("MetapathTracker")
+      .def(torch::init<std::vector<edge_type>,
+                       c10::Dict<rel_type, std::vector<int64_t>>,
+                       std::vector<node_type>>())
+      .def("report_sample_size", &MetapathTracker::report_sample_size)
+      .def("get_sample_size", &MetapathTracker::get_sample_size)
+      .def("init_batch", &MetapathTracker::init_batch)
+      .def("get_neighbor_metapath", &MetapathTracker::get_neighbor_metapath);
```
rusty1s, line 711: drop.

test/classes/test_neighbor_sampler.py
```diff
@@ -50,26 +50,39 @@ def test_hetero_neighbor_sampler_temporal_sample() -> None:
...
+    assert sum(batch['A'] == 0) == 2
+    assert sum(batch['B'] == 0) == 1
+    assert sum(batch['A'] == 1) == 2
+    assert sum(batch['B'] == 1) == 1
+    assert sum(batch['A'] == 2) == 1
+    assert sum(batch['B'] == 2) == 1
+    assert sum(batch['A'] == 3) == 2
+    assert sum(batch['B'] == 3) == 2
+    assert sum(batch['A'] == 4) == 1
+    assert sum(batch['B'] == 4) == 2
```
rusty1s, line 77: short comment on why this is expected would be great.

test/classes/test_neighbor_sampler.py
```diff
@@ -50,26 +50,39 @@ def test_hetero_neighbor_sampler_temporal_sample() -> None:
...
-    assert num_sampled_nodes == {'A': [2, 1, 2], 'B': [1, 2, 0]}
-    assert num_sampled_edges == {'A__to__B': [0, 2, 0], 'B__to__A': [0, 1, 2]}
+    # Due to random shuffle, the output isn't entirely deterministic
+    assert row['A__to__B'].shape[0] == sum(num_sampled_edges['A__to__B'])
+    assert row['B__to__A'].shape[0] == sum(num_sampled_edges['B__to__A'])
+    assert col['A__to__B'].shape[0] == sum(num_sampled_edges['A__to__B'])
+    assert col['B__to__A'].shape[0] == sum(num_sampled_edges['B__to__A'])
+    assert node_id['A'].shape[0] == sum(num_sampled_nodes['A'])
+    assert node_id['B'].shape[0] == sum(num_sampled_nodes['B'])
+    assert sum(batch['A'] == 0) == 2
```
rusty1s, line 68:
```suggestion
    assert (batch['A'] == 0).sum() == 2
```

rusty1s (approved):

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: For each path, we need to track the exact metapath ID because a single node can be reached through different metapaths. I believe that making this a triple is less code mess, however, I'm happy to revisit this later whenever we do a rehaul/cleanup of this codebase.

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: Technically speaking, it's triplet, yeah. Let me rename

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: Good call

pyg_lib/csrc/classes/cpu/neighbor_sampler.cpp
vid-koci: Should have been removed, I think I was clumsy with cleaning.

PR #427, opened 2025-04-10
Fix the Edge Sampler so that it returns Global Edge Ids Instead of Local

zaristei:
I noticed while implementing [this](https://github.com/pyg-team/pytorch_geometric/pull/10126) that the sampler returns unusual values for the row and col fields of edges despite the edge ids being correct. This fixes that issue.
TODO:
Still need to test for edge cases and make sure this doesn't break distributed sampling or the like.


rusty1s:
This doesn't look correct for disjoint sampling, where `node_t` is a pair.

zaristei:
I've been looking through the unit-tests, and it looks like using the local rows and cols is the intended behavior for this package, is that correct?
If so, I don't want to break the pattern. Instead, I can add a flag either here or in `neighbor_sampler.py` in pyg that reconstructs the global row/col vals based on the selected nodes and the local row/col values.

zaristei:
Closing, moving changes to this PR: https://github.com/pyg-team/pytorch_geometric/pull/10200

PR #435, opened 2025-04-22
Build for manylinux

rusty1s:


.github/workflows/building.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, macos-14, windows-2019]
+        os: [ubuntu-latest, ubuntu-latest, macos-latest, windows-latest]
```
vid-koci, line 13: Why is ubuntu repeated here twice?

vid-koci: Also, it's probably good practice to not only include the latest release, but the latest LTS as well - interim releases are not maintained for long (9 months) and hence usually aren't used much.

vid-koci (commented):
I assume that this isn't finished yet since there is a whole lot of commented code

PR #478, opened 2025-06-09
ci: Build manylinux wheels

akihironitta:
Reference: https://cibuildwheel.pypa.io/en/latest/options/


pyproject.toml
```diff
@@ -36,6 +36,25 @@ documentation="https://pyg-lib.readthedocs.io"
...
+    "FORCE_CUDA",
+    "CUDA_HOME",
+    "TORCH_CUDA_ARCH_LIST",
+]
+repair-wheel-command = """
+auditwheel repair \
+  --wheel-dir {dest_dir} {wheel} \
+  --exclude 'libtorch*.so' \
+  --exclude 'libc10*.so' \
+  --exclude 'libgomp*.so*'
```
akihironitta, line 55:
```suggestion
  --exclude 'libtorch*.so' \
  --exclude 'libc10*.so' \
  --exclude 'libgomp*.so*' \
  --exclude 'libcuda.so*'
```

PR #507, opened 2025-08-15
Add AMD GPU ROCm/HIP backend

Looong01:
All test passed. It could be merged.


gpicciuca:
Bumping! Why is there no feedback from the maintainers? No interest in supporting ROCm? Would love to see this merged and have proper wheels released.

gpicciuca:
> > Why is there no feedback from the maintainers? No interest in supporting ROCm?
> 
> @gpicciuca I think we would definitely be happy to add support for ROCm, but we don't have access to AMD GPUs and don't have good ways for us to test ROCm builds at the moment. The number of AMD GPU users is still much smaller than that of NVIDIA GPUs users, and given the fact that we're a very small team, I don't see it feasiable for us to keep maintaining ROCm builds in coming months or year(s) (unless, e.g., AMD team can provide assistance).

I could probably lend a hand here as I own an RX 9070 XT. Using Windows 10 + WSL2 with ROCm 6.4 installed according to AMD's instructions.

Also, have you tried pinging some AMD dev? I think they are active in some Pytorch repository.. so they could potentially contribute to PyG as well i think.. but yea, don't take my words for granted as I don't work for AMD

Looong01:
Sorry, the matrix multiple kernel still need to be implement by composable kernel(ck) library instead of hipblas_lt. It is hard to learn ck.

Looong01:
Have Updated to v0.8.0

PR #542, opened 2025-11-07
Feature/index sort cpp api

jiinyih:
Hi, these are the changes made to be able to import a PyG torchscript model into C++ (Added index_sort.h)

Im not too sure if what done here is correct, but after including this header and some changes to index_sort.cpp, I was able to load a GIN model (given here: https://github.com/pyg-team/pytorch_geometric/blob/master/examples/cpp/README.md) where before it would result in Unknown builtin op: pyg::index_sort.

However, I found that I had to explicitly run pyg::ops::index_sort() before loading the model else it would result in the same error. 

It would be much appreciated if someone could take a look at this.

### DIFF
--- a/pyg_lib/csrc/ops/index_sort.h
+++ b/pyg_lib/csrc/ops/index_sort.h
@@ -7,8 +7,6 @@
 namespace pyg {
 namespace ops {
 
-// Im just following the format given in cpu/index_sort_kernal.cpp
-// and the format of the other headers
 PYG_API std::tuple<at::Tensor, at::Tensor> index_sort(
     const at::Tensor& input,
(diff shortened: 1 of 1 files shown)


pyg_lib/csrc/ops/index_sort.cpp
```diff
@@ -1,9 +1,28 @@
+#include "index_sort.h"
+
 #include <ATen/core/dispatch/Dispatcher.h>
 #include <torch/library.h>
 
 namespace pyg {
 namespace ops {
 
+// I think this is needed as well?
```
rusty1s:
```suggestion
```

PR #638, opened 2026-04-22
Fix incorrect reduction condition in fused_scatter_reduce Triton kernel

Umang-projects:
### Description
While reviewing the Triton implementations for `fused_scatter_reduce`, I noticed a copy-paste error in the `REDUCE1` block of `_fused_scatter_reduce_forward_kernel`. 

Specifically, when `REDUCE1` is evaluated for `min` or `max`, the kernel incorrectly checks `REDUCE2 == 3` and `REDUCE3 == 4` instead of `REDUCE1`. This causes silent mathematical errors and failing reductions when multiple reduction types (including min/max) are passed in specific orders.

### Changes Made:
- Fixed the condition inside `if REDUCE1 > 0:` to correctly check `REDUCE1 == 3` (min) and `REDUCE1 == 4` (max).

*Note: I also noticed the TODOs regarding double computation of `sum/mean`, unrolling the loops (since `tl.static_unroll` is now available), and adding backward passes. I plan to address those optimizations in follow-up PRs once this critical fix is merged.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dropped support for CUDA architectures below `sm_60` (required by cuCollections)
 
 ### Fixed
+
 - Fixed incorrect reduction condition in `fused_scatter_reduce` Triton kernel
 
-### Security 
+### Security
 
 ## [0.6.0] - 2026-03-24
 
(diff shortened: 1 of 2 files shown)


Umang-projects:
Hi @rusty1s ,
I was looking into the Triton kernel implementations and noticed a small bug part in the `fused_scatter_reduce` logic that wa marked as a TODO.
I have fixed the incorrect reduction conditions. Could you please review this when you have some time?

(Note: The Codecov check is failing due to overall target coverage, but the modified lines are fully covered).

Let me know if i need to make any changes.

PR #643, opened 2026-04-29
Add `uv` guard

rusty1s:


.github/actions/setup/action.yml
```diff
@@ -39,6 +39,8 @@ runs:
       shell: bash
 
     - name: Install PyTorch ${{ inputs.torch-version }}+${{ inputs.cuda-version }}
+        env:
```
akihironitta:
```suggestion
      env:
```

akihironitta (approved):

akihironitta:
The remaining failures should be irrelevant to the change in this PR. Reference: https://github.com/pytorch/pytorch/issues/181375

PR #656, opened 2026-05-18
Add `segment_min_coo`

akihironitta:
Tuple-returning min reduction over a sorted COO index. CPU is
deterministic first-match via a sequential E-pass with strict `<`. CUDA
keeps the value pass identical in shape to `segment_sum_coo` (K==1 warp
`SHFL_UP_SYNC` + boundary `atomicMin`; K>1 broadcast template with
`TB ∈ {4, 8, 16, 32}`) but the shuffle propagates ONLY the value — the
argindex is set in a separate post-pass kernel that races to the smallest
matching `e` via `atomicMin` on `int64_t`. Empty buckets reset values
to 0 with the sentinel `arg = src.size(dim)`; backward uses the
`+1`/`narrow` pattern from `scatter_min`.


akihironitta:

PR #657, opened 2026-05-18
Add `segment_max_coo`

akihironitta:
Symmetric to `segment_min_coo`: `>` comparison with
`numeric_limits::lowest()` init. CUDA uses `atomicMax` (commit 5) for the
value pass; arg pass still uses `atomicMin` on argindex for deterministic
first-match on ties. Backward is the same `+1`/`narrow` pattern.


akihironitta:

PR #659, opened 2026-05-18
Add `segment_mean_csr`

akihironitta:
Like `segment_sum_csr`, plus division by row lengths (`indptr.diff()`).
Empty rows get clamped denominators so the output is 0 rather than NaN.
Backward wraps in `if (grad_in.numel() > 0)` like `segment_mean_coo` and
returns `gather_csr(grad_out / count, indptr)`.


akihironitta:

PR #661, opened 2026-05-18
Add `segment_max_csr`

akihironitta:
Symmetric to `segment_min_csr`: `>` comparison with
`numeric_limits::lowest()` init. CUDA uses the same TB=32 one-warp-per-row
shuffle reduction, conditional-swapping `(value, argindex)` pairs when
the partner's value is greater. Same `+1`/`narrow` backward pattern.


akihironitta:

PR #669, opened 2026-06-02
Prepare 0.7.0 release

akihironitta:


.github/workflows/utils/minimal_matrix_linux.json
```diff
@@ -6,7 +6,7 @@
   },
   {
     "torch-version": "2.9.0",
-    "python-version": ["3.10", "3.14"],
+    "python-version": ["3.10", "3.13"],
```
akihironitta, line 9: Although the wheel seems available in their index, the official PyTorch version that started supporting Python 3.14 was PyTorch 2.10.

PR #682, opened 2026-06-08
ci: test Linux aarch64 CPU wheels

babymastodon:
This is an experimental CI change to validate Linux aarch64 CPU wheel builds. It adds an arch entry to the Linux wheel matrix, runs aarch64 jobs on GitHub hosted ARM runners, and leaves the existing x86_64 CUDA matrix unchanged.

CUDA/aarch64 combinations are intentionally out of scope for this PR; those can follow separately once the CPU-only path is validated.

### DIFF
--- a/.github/workflows/_build_linux.yml
+++ b/.github/workflows/_build_linux.yml
@@ -23,7 +23,7 @@ defaults:
 
 jobs:
   wheel:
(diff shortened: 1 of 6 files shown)


akihironitta:
Closing in favor of https://github.com/pyg-team/pyg-lib/pull/683.

